PR #22, opened 2019-06-13
Enable travis and measure code coverage

### LINKED ISSUE #20: Enable travis-ci and measure code coverage

bassosimone:
Closes #20


bassosimone:
:'( apparently the Travis build hook that I have just removed was required to have the decryption of variables, while the Travis app seems less useful. I am so annoyed.

bassosimone:
So, I somehow managed to have Travis work correctly. I suspect this won't last. Still, at least we have something green to show in our README and poor coverage.

I've created this follow-up tickets:

- https://github.com/ooni/probe-engine/issues/24

- https://github.com/ooni/probe-engine/issues/25

@hellais can you please take a look at this branch and provide feedback?

README.md
```diff
@@ -0,0 +1,41 @@
...
+# OONI probe measurement engine
+
+This repository contains OONI probe's [measurement engine](
+https://github.com/ooni/spec/tree/master/probe#engine). That is, the
+piece of software that implements OONI nettests.
+
+## API
+
+You can [browse ooni/probe-engine's API](
+https://godoc.org/github.com/ooni/probe-engine?status.svg)
```
hellais: Are you sure this is the right link? When I click on it I only see an image

README.md
```diff
@@ -0,0 +1,41 @@
...
+at [godoc.org](https://godoc.org/). We don't provide any API
```
hellais: Maybe this should also just link to the godoc of OONI and not the homepage of godoc

README.md
```diff
@@ -0,0 +1,41 @@
...
+stability guarantees and, as such, we will never release v1.0.0
+of this software.
```
hellais: This sounds like a pretty strong statement, perhaps we can just say that currently we don't provide any sort of API stability.

We may still want at some point to cut a 1.0.0 even if it's just to communicate some level of API compatibility for our own purpose.

README.md
```diff
@@ -0,0 +1,41 @@
...
+
+This repository also allows to build [miniooni](cmd/miniooni), a
+small command line client useful to test the functionality in here
+without integrating with OONI probe. You can browse [the manual
+of this tool](
+https://godoc.org/github.com/ooni/probe-engine/cmd/miniooni)
+at [godoc.org](https://godoc.org/). We may change the command line
```
hellais: I think this should just link to the miniooni godoc page or maybe just extend the link to include this string too.

README.md
bassosimone: No, it's wrong. I'm fixing it.

README.md
bassosimone: Yeah, that link to the homepage can be confusing, fixing.

README.md
bassosimone: > perhaps we can just say that currently we don't provide any sort of API stability.

Sounds less strong, okay

README.md
bassosimone: Same as above, won't link to the godoc homepage

PR #33, opened 2019-07-23
session: rename SetAvailableHTTPS* to AddAvailableHTTPS*

### LINKED ISSUE #32: refactor(session): setAvailable... => addAvailable
In session/session.go there is currently a bunch of functions like SetAvailableHTTPSCollector that should be better named AddAvailableHTTPSCollector. In fact, set hints that we're replacing the old value, whereas add hints that we're appending to the existing values list.

xhdix:
closes #32

Files changed:
- session/session.go
- cmd/miniooni/main.go

Renamed functions:
- SetAvailableHTTPSCollector => AddAvailableHTTPSCollector
- SetAvailableHTTPSBouncer => AddAvailableHTTPSBouncer

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -112,16 +112,16 @@ func New(
 	}
 }
 
-// SetAvailableHTTPSBouncer sets the HTTP bouncer base URL.
-func (s *Session) SetAvailableHTTPSBouncer(baseURL string) {
+// AddAvailableHTTPSBouncer adds the HTTP bouncer base URL to the list of URLs that are tried.
+func (s *Session) AddAvailableHTTPSBouncer(baseURL string) {
 	s.AvailableBouncers = append([]model.Service{}, model.Service{
 		Address: baseURL,
 		Type:    "https",
 	})
 }
 
-// SetAvailableHTTPSCollector adds an HTTP bouncer base URL.
-func (s *Session) SetAvailableHTTPSCollector(baseURL string) {
+// AddAvailableHTTPSCollector adds an HTTP collector base URL to the list of URLs that are tried.
+func (s *Session) AddAvailableHTTPSCollector(baseURL string) {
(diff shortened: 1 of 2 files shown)


bassosimone:
This seems to be based off `feature/oonimeasure`. Can it be based off master?

session/session.go
```diff
@@ -112,16 +112,16 @@ func New(
 	}
 }
 
-// SetAvailableHTTPSBouncer sets the HTTP bouncer base URL.
-func (s *Session) SetAvailableHTTPSBouncer(baseURL string) {
+// AddAvailableHTTPSBouncer adds the HTTP bouncer base URL.
```
bassosimone: @xhdix I just noticed that this documentation comment does not make much sense now. Can you make it a bit more specific saying that the URL is added to the list of URLs that are tried?

session/session.go
```diff
@@ -112,16 +112,16 @@ func New(
...
+func (s *Session) AddAvailableHTTPSBouncer(baseURL string) {
 	s.AvailableBouncers = append([]model.Service{}, model.Service{
 		Address: baseURL,
 		Type:    "https",
 	})
 }
 
-// SetAvailableHTTPSCollector adds an HTTP bouncer base URL.
-func (s *Session) SetAvailableHTTPSCollector(baseURL string) {
+// AddAvailableHTTPSCollector adds an HTTP bouncer base URL.
```
bassosimone: @xhdix Like above, plus there's a typo: the comment should mention the collector rather than the bouncer

session/session.go
xhdix: @bassosimone sure! done.

session/session.go
xhdix: @bassosimone done. thanks.

PR #35, opened 2019-07-31
Make minooni logging simpler and more actionable

### LINKED ISSUE #30: miniooni: improve logging
1. display minimal progress on `stdout`

2. disable logging on `stderr` when `-l LOGFILE` is used

3. add timings to all logs

xhdix:
Close #30

### DIFF
--- a/cmd/miniooni/main.go
+++ b/cmd/miniooni/main.go
@@ -10,9 +10,9 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/apex/log"
-	"github.com/apex/log/handlers/multi"
 
 	"github.com/ooni/probe-engine/experiment"
 	"github.com/ooni/probe-engine/experiment/dash"
@@ -39,7 +39,6 @@ type options struct {
 	collectorURL string
 	inputs       []string
 	extraOptions []string
-	logfile      string
 	noBouncer    bool
 	noGeoIP      bool
 	noJSON       bool
@@ -56,6 +55,7 @@ const (
 
 var (
 	globalOptions options
+	startTime     = time.Now()
 )
 
 func init() {
@@ -80,9 +80,6 @@ func init() {
 	getopt.FlagLong(
 		&globalOptions.extraOptions, "option", 'O',
 		"Pass an option to the experiment", "KEY=VALUE",
-	)
-	getopt.FlagLong(
-		&globalOptions.logfile, "logfile", 'l', "Set the logfile path", "PATH",
 	)
 	getopt.FlagLong(
 		&globalOptions.noBouncer, "no-bouncer", 0, "Don't use the OONI bouncer",
@@ -149,7 +146,9 @@ type logHandler struct {
 }
 
 func (h *logHandler) HandleLog(e *log.Entry) (err error) {
-	s := fmt.Sprintf("<%s> %s", e.Level, e.Message)
+	s := fmt.Sprintf("[%14.6f] <%s> %s",
+		time.Since(startTime).Seconds(),
+		e.Level, e.Message)
 	if len(e.Fields) > 0 {
 		s += fmt.Sprintf(": %+v", e.Fields)
 	}
@@ -176,19 +175,11 @@ func main() {
 	if globalOptions.reportfile == "" {
 		globalOptions.reportfile = "report.jsonl"
 	}
-	if globalOptions.logfile != "" {
-		filep, err := os.Create(globalOptions.logfile)
-		if err != nil {
-			log.WithError(err).Fatal("cannot open logfile")
-		}
-		defer filep.Close()
-		logger.Handler = multi.New(logger.Handler, &logHandler{Writer: filep})
-	}
 	workDir, err := os.Getwd()
 	if err != nil {
 		log.WithError(err).Fatal("cannot get current working directory")
 	}
-	log.Log = logger // from now on logs may be redirected
+	log.Log = logger
 
 	ctx := context.Background()
 	var tlsConfig *tls.Config
@@ -214,19 +205,25 @@ func main() {
 	}
 
 	if !globalOptions.noBouncer {
+		log.Info("Looking up OONI backends")
 		if err := sess.MaybeLookupBackends(ctx); err != nil {
 			log.WithError(err).Fatal("cannot lookup OONI backends")
 		}
 	}
 	if !globalOptions.noGeoIP {
+		log.Info("Looking up your location")
 		if err := sess.MaybeLookupLocation(ctx); err != nil {
 			log.WithError(err).Warn("cannot lookup your location")
+		} else {
+			log.Infof("your IP: %s, country: %s, ISP name: %s",
+				sess.Location.ProbeIP, sess.Location.CountryCode, sess.Location.NetworkName)
 		}
 	}
 
 	name := getopt.Args()[0]
 
 	if name == "web_connectivity" {
+		log.Info("Fetching test lists")
 		if len(globalOptions.inputs) <= 0 {
 			list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC(), 100)
 			if err != nil {
@@ -281,7 +278,11 @@ func main() {
 		defer experiment.CloseReport(ctx)
 	}
 
+	inputCount := len(globalOptions.inputs)
+	inputCounter := 0
 	for _, input := range globalOptions.inputs {
+		inputCounter++
+		log.Infof("[%d/%d] running with input: %s", inputCounter, inputCount, input)
 		measurement, err := experiment.Measure(ctx, input)
 		if err != nil {
 			log.WithError(err).Warn("measurement failed")


cmd/miniooni/main.go
```diff
@@ -145,7 +143,20 @@ type logHandler struct {
...
+		if e.Level > log.InfoLevel && !wrong {
+			log.Info(" O_o  Oh! Something is wrong." +
+				" Please try again with the -v option.")
+			wrong = true
+		}
+		return nil
+	}
+	if e.Level > log.InfoLevel && !wrong{
+		wrong = true
+	}
```
bassosimone: Deciding whether there was an error based on the level of logging we see strikes me as weird. I believe the default action on `FatalLevel` is to `exit(1)` anyway, so there is already a clear indication from the Unix shell that something went wrong (and also PowerShell has that). For this reason, I would recommend to remove the part of the implementation dealing with the `wrong` variable.

cmd/miniooni/main.go
```diff
@@ -303,4 +306,11 @@ func main() {
 			}
 		}
 	}
+
+	if wrong {
+		log.Infof("...Finished with error! Duration : %v", time.Since(startTime))
+	}else {
+		log.Infof("...Finished perfectly! Thank you. Duration : %v", time.Since(startTime))
+	}
```
bassosimone: I would strongly suggest to remove the code dealing with `startTime`. Both in the Unix shell and in the PowerShell there are mechanisms to measure the runtime of a command.

xhdix: OK. :D

cmd/miniooni/main.go
```diff
@@ -155,37 +166,29 @@ func (h *logHandler) HandleLog(e *log.Entry) (err error) {
...
 	if len(getopt.Args()) != 1 {
 		log.Fatal("You must specify the name of the experiment to run")
 	}
 	extraOptions := mustMakeMap(globalOptions.extraOptions)
 	annotations := mustMakeMap(globalOptions.annotations)
 
-	logger := &log.Logger{
-		Level:   log.InfoLevel,
-		Handler: &logHandler{Writer: os.Stderr},
-	}
```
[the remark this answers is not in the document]
xhdix: > I would strongly recommend to keep the logger here...

I moved the `logger` because `log.Fatal` was not printed according to our structure! (Duality in displaying logs)

>  ...because you'd be printing before the program has been able to decide whether it has been called with the correct arguments,..

I wrote a simple text, because otherwise it would require more explanation to get started. Like: `web connectivity test started` or something like that. And that required adding more code to the *mini* program. Maybe I can handle this by adding a separate pkg*?

* To handle printing progress percentages(if possible) and more accurate descriptions. 

> (Also, I believe there are better ways to give a sense of progress.)

I have tried different ways to show its progress.
Showing progress on a single line was the worst. All of them use `\r`, which makes the simultaneous display of logs very complicated. (Even deletes the original logs in some cases.)

(Is there a way to get a percentage of progress from MK and other test methods?)

cmd/miniooni/main.go
xhdix: This part helps prevent repetition of the message in `non-verbose` mode:
https://github.com/ooni/probe-engine/blob/c8e30597c66f5d05cb79486450f3ecdf18c8adb0/cmd/miniooni/main.go#L147-L151

This part helps to print the message of success or failure at the end of the program in `verbose` mode:
https://github.com/ooni/probe-engine/blob/c8e30597c66f5d05cb79486450f3ecdf18c8adb0/cmd/miniooni/main.go#L154-L156

Remove?

However, this part `e.Level > log.InfoLevel &&` of this line should be deleted:
https://github.com/ooni/probe-engine/blob/c8e30597c66f5d05cb79486450f3ecdf18c8adb0/cmd/miniooni/main.go#L147-L148

bassosimone: So, if the logger is verbose, you see verbose messages. If it's not, then you don't see them. This is the Unix convention that a program behaves as you wish and does not try to guess what you don't know (e.g. that there are verbose messages you can access if you specify `-v`).

With respect to `wrong`: you cannot decide whether there was a real error based on the log level you see. You can however count the number of failed experiments and used that information.

cmd/miniooni/main.go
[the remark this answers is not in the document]
bassosimone: > [...] because log.Fatal was not printed according to our structure!

Did you test that?

> I wrote a simple text, because otherwise it would require more explanation to get started.

I suggest to write `Running with input ${input}` in the loop way down for every non-empty-string input. That seems much more useful, as a progress estimator, than "starting" because it tells you what input is being processed.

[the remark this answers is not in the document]
bassosimone: > Showing progress on a single line was the worst. All of them use \r, which makes the simultaneous display of logs very complicated. (Even deletes the original logs in some cases.)

Please, just one log line for event. If we ever want to be fancy, we do that in `ooni/probe-cli`, not here.

> (Is there a way to get a percentage of progress from MK and other test methods?)

Yes. You get the `inputs` list length and then compute the percentage as you go. Second thinking about my above suggestion, I guess we can print `[33%] running with input ${input}`.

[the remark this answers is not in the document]
xhdix: > Did you test that?

This line is displayed with the default structure:
https://github.com/ooni/probe-engine/blob/c8e30597c66f5d05cb79486450f3ecdf18c8adb0/cmd/miniooni/main.go#L177

> ..., than "starting" because it tells you what input is being processed.

Good suggestion. But in some (non-verbose) tests, the first message is delayed by more than 5 seconds! No problem if no message is displayed before that?
```
# ./miniooni -no templog.txt web_connectivity 
[      0.000138] <info> Stared...
[     16.723880] <info>  O_o  Oh! Something is wrong. Please try again with the -v option.
[     16.740180] <info> ...Finished with error! Duration : 16.739705905s
```
```
# ./miniooni -no templog.txt web_connectivity 
[      0.000145] <info> Stared...
[     41.552178] <info>  O_o  Oh! Something is wrong. Please try again with the -v option.
[     41.552219] <fatal> cannot lookup OONI backends: map[error:All available bouncers failed]
```

> Please, just one log line for event.

OK! :D 

> Second thinking about my above suggestion 

Yeah! :D,  I had read the messages from bottom to top and answered the same way. :))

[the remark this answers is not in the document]
bassosimone: > This line is displayed with the default structure:

Yeah, it actually should just print without any specific structure.

> the first message is delayed by more than 5 seconds

Then I suggest to print the high level operation being performed (e.g. getting information on available OONI backends)

cmd/miniooni/main.go
```diff
@@ -145,7 +143,15 @@ type logHandler struct {
 }
 
 func (h *logHandler) HandleLog(e *log.Entry) (err error) {
-	s := fmt.Sprintf("<%s> %s", e.Level, e.Message)
+	if e.Level > log.InfoLevel {
+		wrongCounter++
+	}
```
bassosimone: This is not the correct way to count the number of failed tests. To this end, you need to count the number of times running a test returns a non `nil` error. There is no guarantee anywhere that a test that fails will emit just a single log event with log severity being equal to `warning`. What's more, a warning is not necessarily a failure. Hence, this way of counting the number of failures is extremely fragile and should not be used.

cmd/miniooni/main.go
```diff
@@ -145,7 +143,15 @@ type logHandler struct {
...
+	if !globalOptions.verbose && e.Level != log.InfoLevel &&  e.Level != log.FatalLevel {
+		return nil
+	}
```
bassosimone: What are you trying to achieve here? You can set the verbosity of `apex/log` directly and we already do that, so I do not fully understand why you are filtering logs by verbosity also in the handler.

cmd/miniooni/main.go
```diff
@@ -155,6 +161,7 @@ func (h *logHandler) HandleLog(e *log.Entry) (err error) {
 }
 
 func main() {
+
```
bassosimone: Please don't

cmd/miniooni/main.go
```diff
@@ -166,26 +173,17 @@ func main() {
 		Level:   log.InfoLevel,
 		Handler: &logHandler{Writer: os.Stderr},
 	}
+	log.Log = logger // from now on logs may be redirected
```
bassosimone: There is no log redirection on files anymore, so the command can go.

cmd/miniooni/main.go
```diff
@@ -210,11 +208,13 @@ func main() {
 	}
 
 	if !globalOptions.noBouncer {
+		log.Info("Start lookup OONI backends")
```
bassosimone: I suggest `Looking up OONI backends` because you don't print `End lookup...` (and I'd not like this, since it would make the output take more vertical space than required.)

cmd/miniooni/main.go
```diff
@@ -224,13 +224,15 @@ func main() {
...
 	if name == "web_connectivity" {
+		log.Info("Start fetch test lists")
 		list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC())
 		if err != nil {
 			log.WithError(err).Fatal("cannot fetch test lists")
 		}
 		for _, entry := range list {
 			inputs = append(inputs, entry.URL)
 		}
+		log.Infof("{ %v } URLs received",len(inputs))
```
bassosimone: This is not necessary. More details can be deduced using `-v` or from subsequent progress output.

cmd/miniooni/main.go
```diff
@@ -273,7 +275,11 @@ func main() {
 		defer experiment.CloseReport(ctx)
 	}
 
+	step := 100.00 / float64(len(inputs))
+	percent := 0.00
 	for _, input := range inputs {
+		percent += step
+		log.Infof("[%3v%%] running with input { %s }",int(percent),input)
```
bassosimone: Please, make sure you use `go fmt` to apply correct indentation and formatting

cmd/miniooni/main.go
```diff
@@ -273,7 +275,11 @@ func main() {
 		defer experiment.CloseReport(ctx)
 	}
 
+	step := 100.00 / float64(len(inputs))
```
bassosimone: Here you only have weak guarantees that `inputs` is not empty. Please add a comment saying that we know it is not empty because the code above ensures that this is the case.

cmd/miniooni/main.go
bassosimone: I'd also fancy more a `... running with input: %s` format

cmd/miniooni/main.go
```diff
@@ -303,4 +309,11 @@ func main() {
 			}
 		}
 	}
+
+	if wrongCounter > 0 {
+		log.Infof("...Finished with %d errors!", wrongCounter)
```
bassosimone: This information is computed incorrectly, as I mentioned above. Also, I believe it would be more useful to perhaps say the number of failures over the number of inputs. Also, if you structure the code in this case, you don't need to write an `if ... else` construct. You can print stats at the end, no matter what. For example:

```
Running experiment with 100 inputs. Seen 20 failures.
```

cmd/miniooni/main.go
xhdix: OK! But `warning` and `error` levels are a serious issue that should alert a user who never uses `-v`.

cmd/miniooni/main.go
xhdix: `apex/log` does this via `level`. We cannot set the level `info` and prevent it from displaying a `warning`.
https://github.com/apex/log/blob/ff0f66940b829dc66c81dad34746d4349b83eb9e/logger.go#L142
Also, if the user uses `-v`, we should not stop displaying the `warning`.

cmd/miniooni/main.go
xhdix: `log.Log` redirect to `logHandler` with `InfoLevel`

cmd/miniooni/main.go
bassosimone: I believe there is some misunderstanding in how logging should work?

This is how it should work:

1. no `-v` then you see fatal, error, warning and info logs

2. with `-v` you also see debug logs

cmd/miniooni/main.go
bassosimone: > But warning and error levels are a serious issue that should alert a user who never uses -v

As I explained in https://github.com/ooni/probe-engine/pull/35#discussion_r318077908, if we let apex/log do its job, warning and error logs are _always_ shown.

cmd/miniooni/main.go
bassosimone: I don't understand what you mean here. Usually with "redirect" we mean that they are sent to a file, here we're not sending the logs to a file. They're still sent to stderr as with the default logger.

xhdix: We used multi to send logs to file.

cmd/miniooni/main.go
```diff
@@ -166,26 +169,17 @@ func main() {
...
 	if globalOptions.reportfile == "" {
 		globalOptions.reportfile = "report.jsonl"
 	}
-	if globalOptions.logfile != "" {
-		filep, err := os.Create(globalOptions.logfile)
-		if err != nil {
-			log.WithError(err).Fatal("cannot open logfile")
-		}
-		defer filep.Close()
-		logger.Handler = multi.New(logger.Handler, &logHandler{Writer: filep})
```
xhdix: Here!

cmd/miniooni/main.go
bassosimone: right, but that's not the case anymore, so the comment is outdated, no?

cmd/miniooni/main.go
xhdix: >with -v you also see debug logs

with `-v` you also see `all` logs

If you set `level` to `warning` then you see `warning`, `error`, and `fatal` logs

bassosimone: Yes, that's what I mean with "with `-v` you _also_ see debug logs"

xhdix: But if the above line being without `!globalOptions.verbose && ` and if the user uses `-v`, no `warning` and `error` logs will be displayed for him. Because all `log.Logs` have been redirected to this handler.

bassosimone: Because this handle is using `stderr` there is no log redirection. I suggest you try to remove this code, run the program and see whether it's working as intended or not. Maybe I am missing something, but it seems to me these lines are duplicating apex/log functionality and can safely be removed without losing any functionality.

cmd/miniooni/main.go
```diff
@@ -74,11 +74,12 @@ func init() {
...
 	getopt.FlagLong(
-		&globalOptions.extraOptions, "option", 'O',
-		"Pass an option to the experiment", "KEY=VALUE",
+		&globalOptions.inputs, "input", 'i',
+		"Add test-dependent input to the test input", "INPUT",
 	)
 	getopt.FlagLong(
-		&globalOptions.logfile, "logfile", 'l', "Set the logfile path", "PATH",
+		&globalOptions.extraOptions, "option", 'O',
+		"Pass an option to the experiment", "KEY=VALUE",
```
bassosimone: We should try to avoid these artefacts. Also, `extraOptions` comes before alphabetically anyway. Since what we're doing here is just adding the `inputs` flag, via the master branch, let's do this in a way in which we do not create unnecessary noise in the diff. Basically, I am suggesting to move `extraOptions` above `inputs` such that it would not be actually interested by the diff.

cmd/miniooni/main.go
```diff
@@ -210,29 +204,36 @@ func main() {
 	}
 
 	if !globalOptions.noBouncer {
+		log.Info("Looking up OONI backends")
 		if err := sess.MaybeLookupBackends(ctx); err != nil {
 			log.WithError(err).Fatal("cannot lookup OONI backends")
 		}
 	}
 	if !globalOptions.noGeoIP {
+		log.Info("Looking up your location")
```
bassosimone: After having used the program for a while, I think it's better to say something like `... you IP, country, and ISP name`. What do you think?

cmd/miniooni/main.go
```diff
@@ -210,29 +204,36 @@ func main() {
 	}
 
 	if !globalOptions.noBouncer {
+		log.Info("Looking up OONI backends")
```
bassosimone: This log line is rather mysterious. How can we make it more obvious what we are doing?

cmd/miniooni/main.go
```diff
@@ -210,29 +204,36 @@ func main() {
...
-	var inputs []string
 
 	if name == "web_connectivity" {
-		list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC())
-		if err != nil {
-			log.WithError(err).Fatal("cannot fetch test lists")
-		}
-		for _, entry := range list {
-			inputs = append(inputs, entry.URL)
+		log.Info("Fetching test lists")
```
bassosimone: Also this log line could be made less mysterious.

cmd/miniooni/main.go
```diff
@@ -273,20 +274,16 @@ func main() {
 		defer experiment.CloseReport(ctx)
 	}
 
-	for _, input := range inputs {
+	inputCount := len(globalOptions.inputs)
+	inputCounter := 0
+	for _, input := range globalOptions.inputs {
+		inputCounter++
+		log.Infof("[%d/%d] running with input: %s",inputCounter,inputCount,input)
```
bassosimone: Remember to run `go fmt`. You clearly did not :-).

cmd/miniooni/main.go
```diff
@@ -273,20 +274,16 @@ func main() {
...
 			continue
 		}
-		// Remember to omit the user IP.
-		//
-		// This is not ooniprobe and there's no need here to increase the
-		// complexity to make this option configurable. Still, I don't want
-		// this tool to be sharing the IP, because I want to provide the
-		// same default level of sharing of ooniprobe to random people that
-		// may run this tool for development purposes or exploration.
-		measurement.ProbeIP = model.DefaultProbeIP
```
bassosimone: I do not understand where this part of the diff comes from. It seems to me it should not be here?

cmd/miniooni/main.go
```diff
@@ -303,4 +300,5 @@ func main() {
 			}
 		}
 	}
+
```
bassosimone: I'd rather not have an empty line here.

cmd/miniooni/main.go
xhdix: It doesn't give a good data. We can change the word `backends` to something else. :D

bassosimone: Yeah, I agree. What can this something else be?

cmd/miniooni/main.go
xhdix: I ruined everything by one mistake. -_-

cmd/miniooni/main.go
xhdix: I have no idea

cmd/miniooni/main.go
xhdix: This lines has been deleted. I wasn't to blame.

bassosimone: (I was not blaming)

bassosimone:
Thanks!

Issue #28, opened 2019-06-19
Implement and use IP scrubbing safety net


bassosimone:
# Rationale

In MK we traditionally scrub the IP manually from the results JSON for each field where to do so. We plan on doing something similar for Go-native tests. This issue is about creating and implementing a different procedure for achieving the same result that:

1. takes in input a `[]byte` (i.e. the serialised JSON) and a `string` containing the IP address

2. makes sure the `string` actually contains an IP address (using [ParseIP](https://golang.org/pkg/net/#ParseIP))

3. scrubs the IP from the `[]byte` and return the modified `[]byte`

We want to perform IP scrubbing right before sending the JSON to the OONI collector and storing the JSON to disk for the following reasons:

1. safety: if we add another test (or we perform changes) we don’t need to think whether to scrub fields because that’s already handled in the line of code just before submitting and saving the JSON

2. inspectability: the concern is addressed close to where it matters and it’s easier to assess correctness by inspection

I do not expect regressions in terms of performance. However, if don't also perform scrubbing at a more granular level, it seems that may have excessive memory usage, as explained below.

If we only perform scrubbing on the serialised JSON, the memory usage is going to be large in some corner cases. The [Go implementation of Replace](https://golang.org/src/bytes/bytes.go?s=20908:20954#L750) allocates a new large enough slice to write the replacement into. Therefore the memory cost of implementing this operation on the bytes of the JSON is the extra cost caused by allocating and copying the extra bytes that are not part of the body and of the headers. It should also be mentioned that, if the JSON does not contain the IP address, then the memory and copy cost is actually zero.

This suggests that we should scrub in advance all the fields that we can scrub, e.g. the `ProbeIP` field inside the outmost JSON structure, and all other fields that we have identified already as problematic (e.g. HTTP headers and body). Not doing that is probably not smart because it means that, if we measure a large HTTP binary body and, say, we don't scrub the `ProbeIP` in the outmost structure, we need to allocate twice the size for no good reason and we could have avoided that easily by just changing the value of a small field. Considering that we want this code to run on mobile phone, I'd rather have this as a safety net than as the only scrubbing mechanism.

# List of tasks

- [x] understand more in depth what doing this kind of replacement means to address potential concerns regarding security and regarding breaking the JSON

- [ ] make sure that we can pass to MK a flag controlling whether it should or should not perform its own scrubbing procedure (we currently disable scrubbing, as I initially though I could only do it top-level, now I see how this can be problematic)

[...]

PR #37, opened 2019-08-13
Set of changes to unblock ooni/probe-cli#46

bassosimone:
# Rationale

As identified in https://github.com/ooni/probe-cli/pull/46#issuecomment-502572011, the remaining actions before merging https://github.com/ooni/probe-cli/pull/46 are:

> - [ ] lookup the IP address of the resolver in the engine and pass it down to MK (#46 (comment), measurement-kit/measurement-kit#1863) — basically depends on MK v0.10.5
>
> - [ ] make sure IP scrubbing is performed somewhere (it's currently unclear whether it is performed and I'm actually keen on doing it a higher rather than on a lower level); see ooni/probe-engine#28

This PR implements enough functionality in the engine to satisfy the above requirements. We do not implement everything that is described in #28 but we implement enough of it such that the two above criteria are satisfied. This is needed to, among other things, unblock #52.

# Scope

Each commit in this PR could be considered independently, if the overall diff looks too messy. Also, I have added more details to each commit message. The whole diff also has comments.


cmd/miniooni/main.go
```diff
@@ -73,6 +73,10 @@ func init() {
 		&globalOptions.collectorURL, "collector", 'c',
 		"Set collector base URL", "URL",
 	)
+	getopt.FlagLong(
+		&globalOptions.inputs, "input", 'i',
+		"Add test-dependent input to the test input", "INPUT",
+	)
```
bassosimone: This change allowed me to manually test using `ifconfig.io`.

collector/collector.go
```diff
@@ -50,28 +54,100 @@ type ReportTemplate struct {
...
-	ID string `json:"report_id"`
+	ID string
+
+	// client is the client that was used.
+	client *Client
 
-	// Client is the client that was used.
-	Client *Client
+	// privacy contains privacy settings.
+	privacy PrivacySettings
```
bassosimone: I figured that these two fields were public for no reason, so made them private.

collector/collector.go
```diff
@@ -50,28 +54,100 @@ type ReportTemplate struct {
...
 }
 
 // OpenReport opens a new report.
 func (c *Client) OpenReport(
-	ctx context.Context, rt ReportTemplate,
+	ctx context.Context, rt ReportTemplate, ps PrivacySettings,
 ) (*Report, error) {
-	var report Report
+	var or openResponse
```
bassosimone: In the old code, the server could overwrite some fields. This is undesirable. Use a separate struct for getting the result of the JSON API call and then copy the values into the report.

collector/collector.go
```diff
@@ -94,12 +170,16 @@ type updateResponse struct {
 func (r *Report) SubmitMeasurement(
 	ctx context.Context, m *model.Measurement,
 ) error {
+	err := r.privacy.apply(m)
+	if err != nil {
+		return err
+	}
```
bassosimone: This is where we honour the privacy settings

collector/collector.go
```diff
@@ -50,28 +54,100 @@ type ReportTemplate struct {
...
+	// number of cases where we failed to sanitize properly.
+	m.AddAnnotation("_probe_engine_sanitize_test_keys", "true")
+	// We assume that our string replacement here cannot make the resulting
+	// serialized JSON non parseable. Since we are very sure about this, let's
+	// be bold and make the code crash if that happens.
+	if err := json.Unmarshal(data, &m.TestKeys); err != nil {
+		panic(err)
+	}
+	return nil
+}
```
bassosimone: This hunk implements enforcing privacy settings

collector/collector.go
```diff
@@ -50,28 +54,100 @@ type ReportTemplate struct {
...
+	// its job correctly and has not leaked the IP.
+	pip := []byte(m.ProbeIP)
+	if bytes.Count(data, pip) <= 0 {
+		return nil
+	}
+	// Killing the previous TestKeys in my intention reduces the memory
+	// pressure possibly caused by this operation. We have, in fact, two
+	// copies of the `data` slice plus the test keys. TBH, I'm not sure
+	// however that losing all references to the test keys here is enough
+	// that they are immediately cleared, or whether that happens later.
```
bassosimone: I am not sure whether this argument is correct

collector/collector_test.go
```diff
@@ -1,7 +1,10 @@
 package collector_test
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
```
bassosimone: File changed to make sure we scrub as intended in the common case

experiment/experiment.go
```diff
@@ -51,7 +51,8 @@ type Experiment struct {
 }
 
 // New creates a new experiment. You should not call this function directly
-// rather you should do <package>.NewExperiment.
+// rather you should do <package>.NewExperiment where <package> is a subpackage
+// inside of the experiment package, e.g., `.../experiment/ndt7`.
```
bassosimone: Comment was unclear. Fixed it while there.

experiment/experiment.go
```diff
@@ -92,7 +93,7 @@ func (e *Experiment) OpenReport(ctx context.Context) (err error) {
 			TestName:        e.TestName,
 			TestVersion:     e.TestVersion,
 		}
-		e.Report, err = client.OpenReport(ctx, template)
+		e.Report, err = client.OpenReport(ctx, template, e.Session.PrivacySettings)
```
bassosimone: This forwards privacy settings from session to experiment to report.

experiment/experiment.go
```diff
@@ -127,10 +128,7 @@ func (e *Experiment) newMeasurement(input string) model.Measurement {
 	}
 }
 
-// Measure performs a measurement with the specified input. Note that the
-// returned measurement will have all its privacy sensitive values set
-// to their real values. It will be your responsibility to strip them and
-// use the default values if the user told us to do so.
+// Measure performs a measurement with the specified input.
```
bassosimone: I feel better now that we're safe by default rather than depending on people reading the docs.

experiment/ndt7/ndt7.go
```diff
@@ -40,7 +40,7 @@ type TestKeys struct {
 }
 
 func discover(ctx context.Context, sess *session.Session) (string, error) {
-	client := mlabns.NewClient("ndt_ssl", sess.UserAgent())
+	client := mlabns.NewClient("ndt7", sess.UserAgent())
```
bassosimone: This caused `ndt7` to always fail.

experiment/ndt7/ndt7.go
```diff
@@ -69,7 +69,7 @@ func measure(
 	const maxRuntime = 15.0 // second (conservative)
 	testkeys := &TestKeys{}
 	measurement.TestKeys = testkeys
-	client := upstream.NewClient(sess.UserAgent())
+	client := upstream.NewClient(sess.SoftwareName, sess.SoftwareVersion)
```
bassosimone: API changed; cope with that.

go.mod
```diff
@@ -15,11 +15,11 @@ require (
 	github.com/Psiphon-Labs/utls v0.0.0-20181219022742-11a4cc033322 // indirect
 	github.com/Yawning/chacha20 v0.0.0-20170904085104-e3b1f968fc63 // indirect
 	github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412 // indirect
-	github.com/apex/log v1.1.0
```
bassosimone: I've updated all direct dependencies except Psiphon (as documented in the relevant commit).

measurementkit/measurementkit.go
```diff
@@ -69,6 +69,9 @@ type Options struct {
 	// NoGeoIP indicates whether to perform a GeoIP lookup
 	NoGeoIP bool `json:"no_geoip,omitempty"`
 
+	// NoResolverLookup indicates whether to perform a resolver lookup
+	NoResolverLookup bool `json:"no_resolver_lookup"`
+
```
bassosimone: See relevant commit for rationale.

model/model.go
```diff
@@ -78,12 +80,23 @@ type Measurement struct {
 }
 
 // AddAnnotations adds the annotations from input to m.Annotations.
-func (m Measurement) AddAnnotations(input map[string]string) {
+func (m *Measurement) AddAnnotations(input map[string]string) {
+	if m.Annotations == nil {
+		m.Annotations = make(map[string]string)
+	}
```
bassosimone: I noticed in tests that we may panic here, so made this function more robust

model/model.go
```diff
@@ -78,12 +80,23 @@ type Measurement struct {
...
 }
 
+// AddAnnotation adds a single annotations to m.Annotations.
+func (m *Measurement) AddAnnotation(key, value string) {
+	if m.Annotations == nil {
+		m.Annotations = make(map[string]string)
+	}
+	m.Annotations[key] = value
+}
+
```
bassosimone: Added because required by scrubbing code introduced in this PR

model/model.go
```diff
@@ -134,3 +147,6 @@ const (
 	// DefaultResolverIP is the default resolver IP.
 	DefaultResolverIP = "127.0.0.1"
 )
+
+// DefaultProbeASNString is the default probe ASN as a string.
+var DefaultProbeASNString = fmt.Sprintf("AS%d", DefaultProbeASN)
```
bassosimone: For convenience

session/session.go
```diff
@@ -51,6 +52,10 @@ type Session struct {
 	// Location is the probe location.
 	Location *model.LocationInfo
 
+	// PrivacySettings contains the collector privacy settings. The default
+	// is to only redact the user's IP address from results.
+	PrivacySettings collector.PrivacySettings
+
```
bassosimone: Seems a good API to me: you construct session with sane defaults. You can change such defaults. In practice in probe-cli we'll always change `PrivacySettings` values. Note that probe-cli has same names and semantics (i.e. no need to apply any `!` or to map things with different names => less chance of errors.)

bassosimone:
At this point the build is failing because `coveralls` has a boo boo. It's safe to bless as ready for review while I look into that accidental annoyance.

collector/collector.go
```diff
@@ -50,28 +54,100 @@ type ReportTemplate struct {
...
+	}
+	data, err := json.Marshal(m.TestKeys)
+	if err != nil {
+		return err
+	}
+	// The check using Count is to save an unnecessary copy performed by
+	// ReplaceAll when there are no matches into the body. This is what
+	// we would like the common case to be, meaning that the code has done
+	// its job correctly and has not leaked the IP.
+	pip := []byte(m.ProbeIP)
```
hellais: I am not very convinced by this pattern here as it's dependent on the order in which functions are getting called and by inverting the order you run the risk of not scrubbing the right address.

I would be more confident if instead we passed to the `PrivacySettings` structure the `LocationInfo` (https://github.com/ooni/probe-engine/blob/master/model/model.go#L104) struct which is something that we do not need to redact or strip.

This way the scrubbing is done independently from the values inside of the object you are scrubbing.

collector/collector.go
```diff
@@ -50,28 +54,100 @@ type ReportTemplate struct {
...
+	m.TestKeys = nil
+	data = bytes.ReplaceAll(data, pip, []byte(`[REDACTED]`))
+	// We add an annotation such that hopefully later we can measure the
+	// number of cases where we failed to sanitize properly.
+	m.AddAnnotation("_probe_engine_sanitize_test_keys", "true")
+	// We assume that our string replacement here cannot make the resulting
+	// serialized JSON non parseable. Since we are very sure about this, let's
+	// be bold and make the code crash if that happens.
+	if err := json.Unmarshal(data, &m.TestKeys); err != nil {
+		panic(err)
```
hellais: I think it's preferable to return an error in this case rather than doing a panic as that allows the user to decide what is the best course of action (maybe simply not uploading the measurement) instead of leading to a full application crash.

collector/collector.go
hellais: I am pretty sure you don't need to do this. Once the TestKeys value goes out of scope it will get garbage collected.

collector/collector.go
```diff
@@ -50,28 +54,100 @@ type ReportTemplate struct {
...
+	}
+	return nil
+}
+
+type openResponse struct {
+	ID string `json:"report_id"`
```
hellais: The collector may also return an `error` key which you may want to parse as well.

Moreover you may also want to parse the `backend_version` and `supported_formats` keys in the open report message.

session/session.go
```diff
@@ -105,7 +110,11 @@ func New(
 		HTTPNoProxyClient: httpx.NewTracingProxyingClient(
 			logger, nil, tlsConfig,
 		),
-		Logger:          logger,
+		Logger: logger,
+		PrivacySettings: collector.PrivacySettings{
+			IncludeCountry: true,
+			IncludeASN:     true,
+		},
```
hellais: In this constructor is where you could pass a reference to `Location` see: https://github.com/ooni/probe-engine/pull/37/files#r313414579

collector/collector.go
bassosimone: That would break resubmission of a report saved on disk, where `LocationInfo` would be the relative to the resubmitting probe rather than to the measuring probe.

If the concern is that of ordering, which is a wise concern, I can refactor the code such that the concern is completely addressed. Will push a diff in a second.

collector/collector.go
bassosimone: The `error` key is gone in the latest spec. We only mention that:

> Upon receiving a response, new written clients MUST check that the status is 200 before continuing, MUST ensure that the server supports their preferred data submission format, and MUST save the report ID.

I will make the code strictly compliant to the spec.

bassosimone: I re-read the spec and found out I could make it slightly more clear. Please take a look: https://github.com/ooni/spec/pull/153

session/session.go
bassosimone: As I mentioned, I'm not 100% sold on using location because in theory you can run tests with the collector disabled and then attempt to submit later. So, refer to the above discussion regarding how I am planning on dealing with this situation. I am going to close this conversation as we'd otherwise have two redundant branches.

collector/collector.go
hellais: That a good point, I had not considered. I do wonder then if it's wise to do the scrubbing in the moment of submission of the report and rather not do it when you create the measurement itself and write it to disk.

Doing it in the moment of collection also raises the question of how to interpret the privacy settings. I believe the privacy settings of collecting or not the probe IP should be relevant to when the user triggers the test and not when they are uploading them (and I believe this is also what we have been doing historically).

As it is now if the user is in a risky situation and they have disabled to collect their IP, then they re-enable it when they get back home they probably still don't want the measurements from the risky situation to leak their IP even if they later changed the settings in a different context.

bassosimone: So, I think I have found a better pattern that should be robust with respect to the concern you have expressed, and additionally should not break with resubmission.

bassosimone: > . I do wonder then if it's wise to do the scrubbing in the moment of submission of the report [...]

Yes, that's the right thing to do!

bassosimone:
@hellais I believe I have addressed your comments by refactoring the code. PTAL?

model/model.go
```diff
@@ -131,6 +137,63 @@ type LocationInfo struct {
...
+	bpip := []byte(currentIP)
+	if bytes.Count(data, bpip) <= 0 {
+		return nil
+	}
+	// Killing the previous TestKeys in my intention reduces the memory
+	// pressure possibly caused by this operation. We have, in fact, two
+	// copies of the `data` slice plus the test keys. TBH, I'm not sure
+	// however that losing all references to the test keys here is enough
+	// that they are immediately cleared, or whether that happens later.
+	m.TestKeys = nil
```
hellais: As mentioned in the previous round you don't need to do this. The golang garbage collector will reap this object once it goes out of scope (i.e. when you do the assignment below).

Now that I re-read the code I actually think it's a bit dangerous to do this `nil` assignment before you do the `json.Unmarshal` because in case that call fails you have deleted the `TestKeys`.

bassosimone: Fixed in 2550506

bassosimone:
> Is my understanding correct in that you don't record the resolver_ip yet?

That is half correct. We look it up but we don't pass it to MK. It seems actually the right thing to do is to pass that information to MK such that it case use it, as we do for the probe IP.

hellais:
> That is half correct. We look it up but we don't pass it to MK. It seems actually the right thing to do is to pass that information to MK such that it case use it, as we do for the probe IP.

Ok. Does probe-engine write the resolver_ip inside of the `test_keys` once it has looked it up?

bassosimone:
> Ok. Does probe-engine write the resolver_ip inside of the test_keys once it has looked it up?

No, it does not. The main issue is that the `test_keys` are opaque. So, it could in principle parse the `test_keys`, unconditionally overwrite the field and then set it again. I was inclined to let MK do that work, hence https://github.com/measurement-kit/measurement-kit/issues/1883 but now I realise that it's actually more simple to just do what I say above.

bassosimone:
> I was inclined to let MK do that work, hence measurement-kit/measurement-kit#1883 but now I realise that it's actually more simple to just do what I say above.

TBH, I'm not sure anymore of what could be the best approach here. The `resolver_ip` being part of the `test_keys` complicates everything quite a bit. If it's okay for you we can create an issue about solving this problem and I can work on that after this PR is landed?

hellais:
Yes it's fine we can do this as part of another PR.

bassosimone:
> Yes it's fine we can do this as part of another PR.

Okay, issue here https://github.com/ooni/probe-engine/issues/38

bassosimone:
So, today I'll proceed with breaking fixing the build, breaking this down in commits and merging.

bassosimone:
Regarding the Coveralls error, [it turns out yesterday it was down for maintenance](https://github.com/kubeflow/pipelines/issues/1820#issuecomment-520712365).

bassosimone:
I restructured the PR in a set of understandable commits and [fixed a bug that prevented tests without input to run when doing an additional review](https://github.com/ooni/probe-engine/compare/255050625ef064d63a96311b5991a7e061f5f846..861ff039655825b0a16db85e57523e7335b10ef5). I am now going to merge as soon as Travis is green.

PR #39, opened 2019-08-15
Add support for passing the limit and category code to the testlists pkg

hellais:
This is useful so that we can better configure the testing targets.

It's especially important since now the backend will return the full test list by default.

### DIFF
--- a/cmd/miniooni/main.go
+++ b/cmd/miniooni/main.go
@@ -228,7 +228,7 @@ func main() {
 
 	if name == "web_connectivity" {
 		if len(globalOptions.inputs) <= 0 {
-			list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC())
+			list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC(), 100)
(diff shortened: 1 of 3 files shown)


hellais:
Yeah my bad, the tests were not passing for a good reason.

It should now be fixed.

cmd/miniooni/main.go
```diff
@@ -228,7 +228,7 @@ func main() {
 
 	if name == "web_connectivity" {
 		if len(globalOptions.inputs) <= 0 {
-			list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC())
+			list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC(), 100)
```
bassosimone: Can you pass zero here? (I assume zero means infinite)

bassosimone:
Did a second reading and left one single, minor comment.

cmd/miniooni/main.go
bassosimone: On second thought `100` would be fine for now.

Issue #40, opened 2019-08-23
Port OONI Probe tests to golang


hellais:
Port OONI Probe tests to golang

Currently all the OONI Probe tests are implemented inside of MeasurementKit: https://github.com/measurement-kit/measurement-kit/tree/master/src/libmeasurement_kit/nettests.

These are all specified inside of OONI Spec: https://github.com/ooni/spec/tree/master/nettests.

We should start porting these to golang in probe-engine.

The higher priority ones are (sorted somewhat by difficulty) are:
* [ ] [header-field-manipulation](https://github.com/ooni/spec/blob/master/nettests/ts-006-header-field-manipulation.md)
* [ ] [http-invalid-request-line](https://github.com/ooni/spec/blob/master/nettests/ts-007-http-invalid-request-line.md)
* [ ] [whatsapp](https://github.com/ooni/spec/blob/master/nettests/ts-018-whatsapp.md)
* [ ] [facebook-messenger](https://github.com/ooni/spec/blob/master/nettests/ts-019-facebook-messenger.md)
* [ ] [telegram](https://github.com/ooni/spec/blob/master/nettests/ts-020-telegram.md)
* [ ] [web-connectivity](https://github.com/ooni/spec/blob/master/nettests/ts-017-web-connectivity.md) (this is pretty complex)

Moreover the next generation tests should be implemented here like: https://github.com/measurement-kit/measurement-kit/issues?q=is%3Aopen+is%3Aissue+label%3A%22new+test%22

bassosimone:
This is priority/medium meaning that we should do a bit of that in every cycle. This is effort/XL for obvious reasons, i.e., because we need to rewrite all of Measurement Kit in Go :tired_face:.

bassosimone:
We have rewritten all the experiments used by OONI mobile and desktop apps. Yet, we still need to rewrite at least captive-portal and possibly also more experiments. For this reason, this issue should stay open. Work is low prio now.

bassosimone:
The most important experiments have been rewritten in Go. I suggest we take some time during Sprint 21 and evaluate whether we should keep this Epic open or rather create new separate issue for the low priority tests that have not been implemented.

bassosimone:
So, I have determined to keep this issue open to track all the experiments that we may possibly want to implement in the future and that are not implemented now. We are not actively working on this because they are seldom used.

bassosimone:
Since this now only lists extra experiments not used by mobile and desktop OONI apps, I've removed the original milestones.

PR #26, opened 2019-06-16
Start measuring more low-level events

bassosimone:
Related to #12 and #17

PR #41, opened 2019-08-24
Pure Go implementation of Telegram experiment

### LINKED ISSUE #54: telegram: rewrite in Go

bassosimone:
This PR implements (i) TCPConnect and HTTP OONI templates and (ii) the Telegram experiment implemented on top of such functionality. Among other things, this PR is meant to show how one could convert an existing OONI test from Measurement Kit to Go (see #40).

The two main limitations of this code are (1) that we're not taking full advantage of the httpx/httptracex tracer to collect low level information (see #26) and (2) that some keys that the spec says are nullable are emitted as empty strings instead.

The first limitation could be addressed by future commits without changing the Telegram test because what is needed are improvements in the OONI templates layer. The second limitation instead is probably an opportunity to relax the specs. (After all, the Python parser we use should probably use robust coding like `if not value:` rather than `if value is None:` and therefore I believe these changes should not have such a dramatic impact on anything.)

Closes #54

### DIFF
--- a/experiment/ootemplate/http.go
+++ b/experiment/ootemplate/http.go
@@ -211,19 +211,19 @@ func HTTPPerformMany(
 		req = req.WithContext(ctx)
 		requests = append(requests, req)
 	}
+	// TODO(bassosimone): make httptracex able to work in thread safe
+	// fashion. Until then, this code is single threaded.
+	//
+	// Failed build: https://travis-ci.org/ooni/probe-engine/builds/576181138
+	//
+	// Fix: https://github.com/ooni/probe-engine/pull/26/commits/f6f23a718
 	client, measurer := NewHTTPClientWithMeasurer(logger, nil, nil)
-	var waitgroup sync.WaitGroup
-	waitgroup.Add(len(requests))
 	for _, req := range requests {
-		go func(req *http.Request) {
-			resp, err := client.Do(req)
-			if err == nil {
-				ioutil.ReadAll(resp.Body)
-				resp.Body.Close()
-			}
-			waitgroup.Done()
-		}(req)
+		resp, err := client.Do(req)
+		if err == nil {
+			ioutil.ReadAll(resp.Body)
+			resp.Body.Close()
+		}
 	}
-	waitgroup.Wait()
 	return measurer.RoundTrips, nil
 }
--- a/experiment/ootemplate/http.go
+++ b/experiment/ootemplate/http.go
@@ -220,6 +220,8 @@ func HTTPPerformMany(
 	client, measurer := NewHTTPClientWithMeasurer(logger, nil, nil)
 	for _, req := range requests {
 		resp, err := client.Do(req)
+		// Note that here we just drain the body if needed, for correctness, but
+		// otherwise we ignore the result because we have it in measurer.
 		if err == nil {
 			ioutil.ReadAll(resp.Body)
 			resp.Body.Close()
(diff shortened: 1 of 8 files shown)


experiment/ootemplate/http.go
```diff
@@ -0,0 +1,231 @@
...
+) ([]HTTPRoundTrip, error) {
+	var requests []*http.Request
+	for _, t := range templates {
+		req, err := http.NewRequest(t.Method, t.URL, nil)
+		if err != nil {
+			return nil, err
+		}
+		if t.UserAgent == "" {
+			// 11.8% as of August 24, 2019 according to
+			// https://techblog.willshouse.com/2012/01/03/most-common-user-agents/
```
bassosimone: @hellais we probably want to distribute user agents using orchestra, don't we?

hellais: Possibly, though we ought to still have a sane default in case the connection to the orchestra fails.

bassosimone: Yes, sure, I'm chiefly worried that we'll not update the header in a long time.

experiment/telegram/telegram.go
```diff
@@ -18,27 +22,101 @@ const (
...
+				continue
+			}
+			if roundtrip.Failure == "" {
+				testkeys.TelegramHTTPBlocking = false
+				break
+			}
+		}
+		var failure *string
+		count := 0
+		for _, roundtrip := range testkeys.Requests {
```
hellais: You also need to check if the HTML title contains the string "Telegram Web", see: https://github.com/measurement-kit/measurement-kit/blob/master/src/libmeasurement_kit/ooni/telegram.cpp#L101.

experiment/telegram/telegram.go
```diff
@@ -18,27 +22,101 @@ const (
...
+			if !strings.Contains(roundtrip.Request.URL, "web.telegram.org") {
+				continue
```
hellais: Why is this check being done to ensure the URL contains the string web.telegram.org?

bassosimone:
We should resume working on this PR with https://github.com/ooni/probe-engine/issues/68 in mind.

experiment/telegram/telegram.go
bassosimone: Thank you, you're right. I ported this change to #76.

experiment/telegram/telegram.go
bassosimone: Because the HTTP results include both web and non-web endpoints and they are used to compute different sets of output variables. Hence, in this specific case, I am making sure I am not using web endpoints because they should not be used to compute this stat.

bassosimone:
> I like how tests are structured and they appear pretty readable and compact.
>
> However the logic for determining the block should be aligned to the current implementation as mentioned in a comment.

Thanks, @hellais. I believe I will not move forward the work in here, because #76 is a better implementation of this concept where we do less work to obtain same results.

I have forward ported all your requests here to #76. I will keep this branch open for a little more until  we start to converge with #76 and then close it.

experiment/telegram/telegram.go
hellais: It's probably a good idea to mention this in a comment, it was not immediate to me that was the reason.

bassosimone: Yeah, I did that here: https://github.com/ooni/probe-engine/pull/76/files#diff-57ae86adcc9a6f48b8959ad307776b49R91

bassosimone:
Superseded by https://github.com/ooni/probe-engine/pull/103

PR #50, opened 2019-09-07
Add example experiment

### LINKED ISSUE #36: Write example experiment
To demonstrate how to use the experiment API it would be useful to have a test which is a "hello world" that basically just writes some values inside of the `test_keys`.

Here is an example of such hello world test inside of probe-legacy: https://github.com/ooni/probe-legacy/blob/master/ooni/nettests/examples/example_simple.py

xhdix:
close #36

### DIFF
--- a/cmd/miniooni/main.go
+++ b/cmd/miniooni/main.go
@@ -10,12 +10,14 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/apex/log"
 	"github.com/apex/log/handlers/multi"
 
 	"github.com/ooni/probe-engine/experiment"
 	"github.com/ooni/probe-engine/experiment/dash"
+	"github.com/ooni/probe-engine/experiment/example"
 	"github.com/ooni/probe-engine/experiment/fbmessenger"
 	"github.com/ooni/probe-engine/experiment/hhfm"
 	"github.com/ooni/probe-engine/experiment/hirl"
@@ -270,6 +272,8 @@ func main() {
 		experiment = web_connectivity.NewExperiment(sess, web_connectivity.Config{})
 	} else if name == "whatsapp" {
 		experiment = whatsapp.NewExperiment(sess, whatsapp.Config{})
+	} else if name == "example" {
+		experiment = example.NewExperiment(sess, example.Config{2 * time.Second})
 	} else {
 		log.Fatalf("Unknown experiment: %s", name)
 	}
--- a/experiment/example/example.go
+++ b/experiment/example/example.go
@@ -1,0 +1,68 @@
+// Package example contains a simple example of experiment.
+package example
+
+import (
+	"context"
+	"time"
+
+	"github.com/ooni/probe-engine/experiment"
+	"github.com/ooni/probe-engine/experiment/handler"
+	"github.com/ooni/probe-engine/model"
+	"github.com/ooni/probe-engine/session"
+)
+
+const (
+	testName    = "example"
+	testVersion = "0.0.1"
+)
+
+// Config contains the experiment config.
+//
+// This contains all the settings that user can set to modify the behaviour
+// of this experiment.
+type Config struct {
+	SleepTime time.Duration
+}
+
+// TestKeys contains the experiment's result.
+//
+// This is what will end up into the Measurement.TestKeys field
+// when you run this experiment.
+// In other words, the variables in this struct will be
+// the particular results of this experiment.
+type TestKeys struct {
+	Success bool `json:"success"`
+}
+
+// measure is the main function of each experiment,
+// that is called by the NewExperiment.
+func measure(
+	ctx context.Context, sess *session.Session, measurement *model.Measurement,
+	callbacks handler.Callbacks, config Config,
+) error {
+	testkeys := &TestKeys{Success: true}
+	measurement.TestKeys = testkeys
+	//let's pretend to do something..
+	time.Sleep(config.SleepTime)
+	return nil
+}
+
+// NewExperiment creates a new experiment.
+//
+// This is the function that you call to create an instance of this experiment.
+// Once you have created an instance, you can use directly the
+// generic experiment API.
+func NewExperiment(
+	sess *session.Session, config Config,
+) *experiment.Experiment {
+	return experiment.New(
+		sess, testName, testVersion,
+		func(
+			ctx context.Context,
+			sess *session.Session,
+			measurement *model.Measurement,
+			callbacks handler.Callbacks,
+		) error {
+			return measure(ctx, sess, measurement, callbacks, config)
+		})
+}
--- a/experiment/example/example_test.go
+++ b/experiment/example/example_test.go
@@ -1,0 +1,47 @@
+package example_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/apex/log"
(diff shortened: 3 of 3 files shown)


experiment/example/example.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+const (
+	testName    = "example"
+	testVersion = "0.0.1"
+)
+
+// Config contains the experiment config.
+type Config struct {
+	sleepTime time.Duration
```
bassosimone: We want this field to be visible outside this package, hence it must be `SleepTime`.

experiment/example/example.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// TestKeys contains the experiment's result.
+//
+// This is what will end up into the Measurement.TestKeys field
+// when you run this experiment.
+type TestKeys struct {
+	foo     string `json:"foo"`
+	success bool   `json:"success"`
```
bassosimone: I believe it suffices to have a single, user-visible key here. I'd say we'll keep `Success`.

experiment/example/example.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func measure(
+	ctx context.Context, sess *session.Session, measurement *model.Measurement,
+	callbacks handler.Callbacks, config Config,
+) error {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
```
bassosimone: The two above lines are not required, because we don't use the context here.

experiment/example/example.go
```diff
@@ -0,0 +1,68 @@
...
+	config.sleepTime = 10 * time.Second
```
bassosimone: I'd rather just use the user-provided value.

experiment/example/example.go
```diff
@@ -0,0 +1,68 @@
...
+	testkeys := &TestKeys{
+		foo:     "bar",
+		success: false,
```
bassosimone: I think it's nicer to set success to `true`.

experiment/example/example.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	measurement.TestKeys = testkeys
+	dosomething(config)
```
bassosimone: I think it's fine to sleep directly here without deferring to a secondary function

experiment/example/example.go
```diff
@@ -0,0 +1,68 @@
...
+	testkeys.success = true
```
bassosimone: This is not required per my comment above.

experiment/example/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+const (
+	softwareName    = "ooniprobe-example"
+	softwareVersion = "0.0.1"
+)
+
+func TestIntegration(t *testing.T) {
+	if !measurementkit.Available() {
+		t.Skip("Measurement Kit not available; skipping")
+	}
```
bassosimone: We can safely skip the three lines above, since this experiment does not use MK

experiment/example/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+		log.Log, softwareName, softwareVersion, "../../testdata", nil, nil,
+	)
+	if err := sess.MaybeLookupBackends(ctx); err != nil {
+		t.Fatal(err)
+	}
+	if err := sess.MaybeLookupLocation(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	experiment := example.NewExperiment(sess, example.Config{})
```
bassosimone: Here is probably the place where to configure the sleep time.

experiment/example/example.go
bassosimone: @xhdix I think also `Success` should be visible outside this package

experiment/example/example.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const (
+	testName    = "example"
+	testVersion = "0.0.1"
+)
+
+// Config contains the experiment config.
+//
+//This includes all variables that will be used inside or outside the module.
```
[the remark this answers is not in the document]
bassosimone: (Also: note that _module_ is Python's terminology, _package_ is Go's terminology)

experiment/example/example.go
```diff
@@ -0,0 +1,63 @@
...
+	callbacks handler.Callbacks, config Config,
+) error {
+	testkeys := &TestKeys{Success: true}
+	measurement.TestKeys = testkeys
+	time.Sleep(config.SleepTime)
+	return nil
+}
+
+// NewExperiment creates a new experiment.
+// This is the preliminary function that is called from outside the module.
```
bassosimone: Maybe "This is the function that you call to create an instance of this experiment. Once you have created an instance, you can use directly the generic experiment API."

xhdix: Although this comment is still a little difficult to understand, it's better than my comment. :P

experiment/example/example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	)
+	if err := sess.MaybeLookupBackends(ctx); err != nil {
+		t.Fatal(err)
+	}
+	if err := sess.MaybeLookupLocation(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	experiment := example.NewExperiment(
+		sess, example.Config{SleepTime: 10 * time.Second})
```
xhdix: In the next changes I will take parenthesis to the next line.
(In this case it will not be out of structure)

experiment/example/example.go
bassosimone: 🤣

xhdix:
> I have a sense that it's useful to also wire in this experiment in the `miniooni` command line, to show also how that can be done. Do you think you can also do that as part of this PR?

Yes, I did, but I didn't add it.
done.
Thanks for your comments.

cmd/miniooni/main.go
```diff
@@ -270,6 +272,8 @@ func main() {
 		experiment = web_connectivity.NewExperiment(sess, web_connectivity.Config{})
 	} else if name == "whatsapp" {
 		experiment = whatsapp.NewExperiment(sess, whatsapp.Config{})
+	} else if name == "example" {
+		experiment = example.NewExperiment(sess, example.Config{10 * time.Second})
```
bassosimone: I suggest to use `1` second as the timeout because `10` seconds is probably annoying :-D

xhdix: Maybe `2` second?  (`time.Second` == `1s`) 
Although this is an example. :D

experiment/example/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	)
+	if err := sess.MaybeLookupBackends(ctx); err != nil {
+		t.Fatal(err)
+	}
+	if err := sess.MaybeLookupLocation(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	experiment := example.NewExperiment(
+		sess, example.Config{SleepTime: 10 * time.Second},
```
xhdix: What about this line?

cmd/miniooni/main.go
bassosimone: Sure, also two seconds is fine

bassosimone:
Thanks, I will merge tomorrow!

bassosimone:
(Did run tests locally, they were all green)

PR #59, opened 2019-09-12
fix(collector): specify format and data-format-version

bassosimone:
Spotted by @FedericoCeratto (thanks!)

"This could have been avoided with end-to-end testing" 🤦

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -15,6 +15,14 @@ import (
 	"github.com/ooni/probe-engine/model"
 )
 
+const (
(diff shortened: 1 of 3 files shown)


bassosimone:
# TODO list for continuing this PR

- [ ] make sure that keys are sorted alphabetically

- [ ] document this as a SHOULD in the spec

- [ ] write a functional test where we verify we get the JSON we expect

(as discussed in chat with @FedericoCeratto)

PR #61, opened 2019-09-13
dash: rewrite using neubot/dash

### LINKED ISSUE #51: dash: rewrite using github.com/neubot/dash
I've recently written a Go implementation of Neubot's DASH test at github.com/neubot/dash. This issue is about performing the following tasks:

- [ ] replace the current implementation with code calling github.com/neubot/dash directly

- [ ] make sure that the resulting JSON is consistent with what you obtain using MK and, if not, take the proper actions to make sure it becomes consistent

xhdix:
close #51
(I'm not sure whether the codes is correct or not.)

### DIFF
--- a/experiment/dash/dash_otherwise.go
+++ b/experiment/dash/dash_otherwise.go
@@ -1,0 +1,154 @@
+// +build !cgo
+
+// Package dash contains the dash network experiment. This file
+// in particular is a pure-Go implementation of that.
+package dash
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/dustin/go-humanize"
+	"github.com/montanaflynn/stats"
+	"github.com/neubot/dash/client"
+	neubotModel "github.com/neubot/dash/model"
+	"github.com/ooni/probe-engine/experiment"
+	"github.com/ooni/probe-engine/experiment/handler"
+	"github.com/ooni/probe-engine/model"
+	"github.com/ooni/probe-engine/session"
+)
+
+const (
+	testName       = "dash"
+	testVersion    = "0.8.0"
+	defaultTimeout = 55 * time.Second
+	totalStep      = 15.0
+)
+
+// Config contains the experiment config.
+type Config struct{}
+
+// Simple contains the experiment total summary
+type Simple struct {
+	ConnectLatency  float64 `json:"connect_latency"`
+	MedianBitrate   int64   `json:"median_bitrate"`
+	MinPlayoutDelay float64 `json:"min_playout_delay"`
+}
+
+// TestKeys contains the test keys
+type TestKeys struct {
+	Simple Simple `json:"simple"`
+
+	// Failure is the failure string
+	Failure string `json:"failure"`
+
+	ReceiverData []neubotModel.ClientResults `json:"receiver_data"`
+}
+
+func measure(
+	ctx context.Context, sess *session.Session, measurement *model.Measurement,
+	callbacks handler.Callbacks, config Config,
+) error {
+	testkeys := &TestKeys{}
+	measurement.TestKeys = testkeys
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+	client := client.New(sess.SoftwareName, sess.SoftwareVersion)
+	// StartDownload starts the DASH download. It returns a channel where
+	// client measurements are posted, or an error.
+	ch, err := client.StartDownload(ctx)
+	if err != nil {
+		testkeys.Failure = err.Error()
+		return err
+	}
+	callbacks.OnProgress(0, fmt.Sprintf("server: %s", client.FQDN))
+	var rates []float64
+	var frameReadyTime float64 = 0.0
+	var playTime float64 = 0.0
+	percentage := 0.0
+	step := 100 / (totalStep + 1) / 100
+	for results := range ch {
+		percentage += step
+		message := fmt.Sprintf(
+			"rate: %s/s speed: %s/s elapsed: %.2f s",
+			humanize.Bytes(uint64(results.Rate*1000/8)), // Rate is kbit in SI size
+			humanize.Bytes(uint64(float64(results.Received)/results.Elapsed)),
+			results.Elapsed,
+		)
+		callbacks.OnProgress(percentage, message)
+		data, err := json.Marshal(results)
+		if err != nil {
+			testkeys.Failure = err.Error()
+			return err
+		}
+		sess.Logger.Debugf("%s", string(data))
+
+		// Here we're computing stats inline
+		// rates is used to calculate MedianBitrate.
+		rates = append(rates, float64(results.Rate))
+
+		if testkeys.Simple.ConnectLatency == 0.0 {
+			// It is always equal for all the records
+			testkeys.Simple.ConnectLatency = results.ConnectTime
+		}
+
+		frameReadyTime += float64(results.Elapsed)
+		if playTime == 0 {
+			playTime += frameReadyTime
+		} else {
+			playTime += float64(results.ElapsedTarget)
+		}
+		var stall float64 = frameReadyTime - playTime
+		if stall > testkeys.Simple.MinPlayoutDelay {
+			testkeys.Simple.MinPlayoutDelay = stall
+		}
+		testkeys.ReceiverData = append(testkeys.ReceiverData, results)
+	}
+
+	median, err := stats.Median(rates)
+	testkeys.Simple.MedianBitrate = int64(median)
+	if err != nil {
+		testkeys.Failure = err.Error()
+		return err
+	}
+	if client.Error() != nil {
+		testkeys.Failure = err.Error()
+		return client.Error()
+	}
+	data, err := json.Marshal(client.ServerResults())
+	if err != nil {
+		testkeys.Failure = err.Error()
+		return err
+	}
+	sess.Logger.Debugf("Server result: %s", string(data))
+	callbacks.OnProgress(1, "done")
+	sess.Logger.Debugf("Test Summary: ")
+	sess.Logger.Debugf("Connect latency: %s",
+		// convert to nanoseconds
+		time.Duration(testkeys.Simple.ConnectLatency*1000000000),
+	)
+	sess.Logger.Debugf("Median bitrate: %s/s",
+		// MedianBitrate is kbit in SI size
+		humanize.IBytes(uint64(testkeys.Simple.MedianBitrate*1000/8)),
+	)
+	sess.Logger.Debugf("Min. playout delay: %.3f s", testkeys.Simple.MinPlayoutDelay)
+	return nil
+}
+
+// NewExperiment creates a new experiment.
+func NewExperiment(
+	sess *session.Session, config Config,
+) *experiment.Experiment {
+	return experiment.New(
+		sess, testName, testVersion,
+		func(
+			ctx context.Context,
+			sess *session.Session,
+			measurement *model.Measurement,
+			callbacks handler.Callbacks,
+		) error {
+			return measure(ctx, sess, measurement, callbacks, config)
+		})
+}
--- a/experiment/dash/dash_cgo.go
+++ b/experiment/dash/dash_cgo.go
@@ -1,3 +1,5 @@
+// +build cgo
+
 // Package dash contains the dash network experiment.
 package dash
 
--- a/experiment/dash/dash_cgo_test.go
+++ b/experiment/dash/dash_cgo_test.go
@@ -1,0 +1,50 @@
+// +build cgo
+
+package dash_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/apex/log"
+	"github.com/ooni/probe-engine/experiment/dash"
+	"github.com/ooni/probe-engine/measurementkit"
+	"github.com/ooni/probe-engine/session"
+)
+
+const (
+	softwareName    = "ooniprobe-example"
+	softwareVersion = "0.0.1"
+)
+
+func TestIntegration(t *testing.T) {
+	if !measurementkit.Available() {
+		t.Skip("Measurement Kit not available; skipping")
+	}
+	log.SetLevel(log.DebugLevel)
+	ctx := context.Background()
+
+	sess := session.New(
+		log.Log, softwareName, softwareVersion, "../../testdata", nil, nil,
+	)
+	if err := sess.MaybeLookupBackends(ctx); err != nil {
+		t.Fatal(err)
+	}
+	if err := sess.MaybeLookupLocation(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	experiment := dash.NewExperiment(sess, dash.Config{})
+	if err := experiment.OpenReport(ctx); err != nil {
+		t.Fatal(err)
+	}
+	defer experiment.CloseReport(ctx)
+
+	measurement, err := experiment.Measure(ctx, "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := experiment.SubmitMeasurement(ctx, &measurement); err != nil {
+		t.Fatal(err)
+	}
+}
--- a/experiment/dash/dash_otherwise_test.go
+++ b/experiment/dash/dash_otherwise_test.go
@@ -1,3 +1,5 @@
+// +build !cgo
+
 package dash_test
 
 import (


experiment/dash/dash.go
```diff
@@ -3,42 +3,131 @@ package dash
...
+
+type Simple struct {
+	Connect_latency   float64 `json:"connect_latency"`
+	Median_bitrate    float64 `json:"median_bitrate"` // median_bitrate is int in MK
+	Min_playout_delay float64 `json:"min_playout_delay"`
+}
+
+// TestKeys contains the test keys
+type TestKeys struct {
+	Simple Simple `json:"simple"`
```
xhdix:
```suggestion
	Simple `json:"simple"`
```

experiment/dash/dash.go
```diff
@@ -3,42 +3,131 @@ package dash
...
 
 // Config contains the experiment config.
-type Config struct {
-	// LogLevel is the MK log level. Empty implies "WARNING".
-	LogLevel string
+type Config struct{}
+
+type Simple struct {
+	Connect_latency   float64 `json:"connect_latency"`
+	Median_bitrate    float64 `json:"median_bitrate"` // median_bitrate is int in MK
```
xhdix:
```suggestion
	Median_bitrate    int64 `json:"median_bitrate"`
```

experiment/dash/dash.go
bassosimone: What is that?

xhdix: Nothing. There is no different. Can be deleted. : D

experiment/dash/dash.go
```diff
@@ -3,42 +3,131 @@ package dash
...
+		Simple: Simple{
+			Connect_latency:   0.0,
+			Min_playout_delay: 0.0,
+		},
+	}
+	measurement.TestKeys = testkeys
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+	client := client.New(sess.SoftwareName, sess.SoftwareVersion)
+	ch, err := client.StartDownload(ctx)
```
xhdix: I think the name "ch" should be changed to something better.

experiment/dash/dash.go
```diff
@@ -3,42 +3,131 @@ package dash
...
+	if err != nil {
+		testkeys.Failure = err.Error()
+		return err
+	}
+	sess.Logger.Debugf("Server result: %s", string(data))
+	callbacks.OnProgress(1, "done")
+	sess.Logger.Info("Test Summary: ")
+	sess.Logger.Infof("Connect latency: %s",
+		// convert to nanoseconds
+		time.Duration(testkeys.Simple.Connect_latency*1000000000),
```
xhdix: (The situation here is unclear. We have no value yet.)

experiment/dash/dash.go
xhdix:
```suggestion
	channel, err := client.StartDownload(ctx)
```

experiment/dash/dash.go
bassosimone: I'm puzzled by this change. Why is it a good idea?

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+		Simple: Simple{
+			Connect_latency:   0.0,
+			Min_playout_delay: 0.0,
+		},
+	}
+	measurement.TestKeys = testkeys
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+	client := client.New(sess.SoftwareName, sess.SoftwareVersion)
+	channel, err := client.StartDownload(ctx)
```
bassosimone: This codebase calls a channel `ch` unless there's more than one channel. Please, follow this convention.

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
 	"github.com/ooni/probe-engine/session"
 )
 
 const (
-	testName    = "dash"
-	testVersion = "0.7.0"
+	testName       = "dash"
+	testVersion    = "0.8.0"
+	defaultTimeout = 55 * time.Second
+	maxRuntime     = 15.0
```
bassosimone: Dash runs for 30 seconds in many cases. I'd say 30 here.

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
 )
 
 // Config contains the experiment config.
-type Config struct {
-	// LogLevel is the MK log level. Empty implies "WARNING".
-	LogLevel string
+type Config struct{}
+
+type Simple struct {
```
bassosimone: Please document the structure

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+	Connect_latency   float64 `json:"connect_latency"`
```
bassosimone: `Connect_latency` is not idiomatic. Please, use `ConnectLatency`.

bassosimone: (Go uses CamelCase. Using snake_case, or Snake_case, or Snake_Case, is very uncommon.)

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+	Median_bitrate    int64   `json:"median_bitrate"`
+	Min_playout_delay float64 `json:"min_playout_delay"`
+}
+
+// TestKeys contains the test keys
+type TestKeys struct {
+	Simple `json:"simple"`
```
bassosimone: Can we avoid using embedding here? Why embedding only that structure and not others? I think that's confusing.

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+
+	Client_resolver string `json:"client_resolver"`
+
+	// Failure is the failure string
+	Failure string `json:"failure"`
+
+	Receiver_data []neubotModel.ClientResults `json:"receiver_data"`
```
bassosimone: Please, run `go vet`, install `go lint` and run `go lint`, on this specific file.

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
-	settings := measurementkit.NewSettings(
-		"Dash", sess.SoftwareName, sess.SoftwareVersion,
-		sess.CABundlePath(), sess.ProbeASNString(), sess.ProbeCC(),
-		sess.ProbeIP(), sess.ProbeNetworkName(), config.LogLevel,
-	)
-	out, err := measurementkit.StartEx(settings, sess.Logger)
+	testkeys := &TestKeys{
+		Simple: Simple{
+			Connect_latency:   0.0,
+			Min_playout_delay: 0.0,
```
bassosimone: Structures are zero initialised by default. You don't need doing this.

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+	channel, err := client.StartDownload(ctx)
 	if err != nil {
+		testkeys.Failure = err.Error()
 		return err
 	}
-	for ev := range out {
-		mkevent.Handle(sess, measurement, ev, callbacks)
+	callbacks.OnProgress(0, fmt.Sprintf("server: %s", client.FQDN))
+	var rates []float64
+	var frame_ready_time float64 = 0.0
```
bassosimone: (Please, snakeCase)

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+	var play_time float64 = 0.0
+	percentage := 0.00
+	step := 100 / (maxRuntime + 1) / 100
+	for results := range channel {
+		percentage += step
+		message := fmt.Sprintf(
+			"rate: %d kbit/s speed: %.2f kbit/s elapsed: %.2f s",
+			results.Rate,
+			float64(results.Received*8/1024)/results.Elapsed,
```
bassosimone: (We are not married to using `kbit/s` always. Maybe we can humanise?

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+			testkeys.Failure = err.Error()
+			return err
+		}
+		sess.Logger.Debugf("%s", string(data))
+
+		rates = append(rates, float64(results.Rate))
+		if testkeys.Simple.Connect_latency == 0.0 {
+			// It is always equal for all the records
+			testkeys.Simple.Connect_latency = results.ConnectTime
+		}
```
bassosimone: We want to compute stats at the end. Server provided stats may be useful. (For example, if we have a better measurement of the RTT provided by the server, we want to use it.)

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+		)
+		callbacks.OnProgress(percentage, message)
+		data, err := json.Marshal(results)
+		if err != nil {
+			testkeys.Failure = err.Error()
+			return err
+		}
+		sess.Logger.Debugf("%s", string(data))
+
+		rates = append(rates, float64(results.Rate))
```
bassosimone: I was initially skeptical about computing stats inline, but that may actually be better than doing it at the end. Can I suggest to add a comment saying that here we're computing stats inline?

experiment/dash/dash.go
```diff
@@ -3,42 +3,132 @@ package dash
...
+	}
+	data, err := json.Marshal(client.ServerResults())
+	if err != nil {
+		testkeys.Failure = err.Error()
+		return err
+	}
+	sess.Logger.Debugf("Server result: %s", string(data))
+	callbacks.OnProgress(1, "done")
+	testkeys.Client_resolver = sess.Location.ResolverIP
+	sess.Logger.Info("Test Summary: ")
```
bassosimone: Printing summary at info level is an activity reserved to `github.com/ooni/probe-cli`. We should not use the info level in this codebase, because this is part of the (currently implicit) contract with the real OONI client.

experiment/dash/dash.go
xhdix: It's just not common.
https://github.com/google/gopacket/blob/2d7fab0d91d6bb77c1df6bdeb222270a2aa13820/layers/tcp.go#L21

experiment/dash/dash.go
xhdix: I think `ch` is a little confusing for someone who wants to look at code for the first time. So should a good explanation be written?

experiment/dash/dash.go
xhdix: This is the number of requests (`ch`). I think it's always 15 times.

experiment/dash/dash.go
xhdix: At first I only followed the rules of MK, so it was easier to refer back during review.

experiment/dash/dash.go
xhdix: This is the structure that was in MK. Doesn't that change cause problems for those using the API? 
https://explorer.ooni.org/measurement/20190913T235806Z_AS266347_h70ebKc0sVX7gz4TusAIpjfWeQjzUS4j2ZB6SPdT7yfZtIMASu

experiment/dash/dash.go
xhdix: Nothing! Do you think something is wrong here?

experiment/dash/dash.go
xhdix: I don't remember where I imitated `maxRuntime`.  :))

bassosimone:
We should resume working on this PR with https://github.com/ooni/probe-engine/issues/68 in mind.

experiment/dash/dash.go
bassosimone: I can understand that, but I'd say writing idiomatic code is also important!

experiment/dash/dash.go
bassosimone: I just mean you should write `Simple Simple `json:"simple"`

experiment/dash/dash.go
bassosimone: No, it was just a reminder. (If you use VSCode with the Go plugin, every time you save a file, it is reformatted, linted, and vetted.)

experiment/dash/dash.go
bassosimone: ofc, I meant camelCase

experiment/dash/dash.go
```diff
@@ -3,42 +3,134 @@ package dash
...
 
 // Config contains the experiment config.
-type Config struct {
-	// LogLevel is the MK log level. Empty implies "WARNING".
-	LogLevel string
+type Config struct{}
+
+// Simple contains the experiment total summary
+type Simple struct {
+	ConnectLatency  float64 `json:"Connect_latency"`
```
bassosimone: Nit: it should be `connect_latency` inside of `json:"..."`

experiment/dash/dash.go
```diff
@@ -3,42 +3,134 @@ package dash
...
+	MedianBitrate   int64   `json:"Median_bitrate"`
```
bassosimone: Nit: `"median_bitrate"`

experiment/dash/dash.go
```diff
@@ -3,42 +3,134 @@ package dash
...
+	var frameReadyTime float64 = 0.0
+	var playTime float64 = 0.0
+	percentage := 0.0
+	step := 100 / (totalStep + 1) / 100
+	for results := range ch {
+		percentage += step
+		message := fmt.Sprintf(
+			"rate: %s/s speed: %s/s elapsed: %.2f s",
+			humanize.IBytes(uint64(results.Rate*1000/8)), // Rate is kbit in SI size
+			humanize.IBytes(uint64(float64(results.Received)/results.Elapsed)),
```
bassosimone: I am always undecided whether to print `KiB/s` or `kbit/s`. Curious: why did you choose the former?

experiment/dash/dash.go
bassosimone: To word this differently: I believe the code computing stats is less obvious than the code above. It would therefore be beneficial to introduce a couple of comments says what it is doing and why.

experiment/dash/dash.go
xhdix: For this I had mentioned somewhere this line:
https://github.com/google/gopacket/blob/33810c487ced200c983df5c45e3ebd2adc404b71/layers/tcp.go#L21
If you think "double `Simple`" is better. I'll do it.

bassosimone: Yes, please, I'd find a single `Simple` confusing. My understanding is that embedding is Go's way to inherit from another "class" and I don't want to give this impression here.

experiment/dash/dash.go
xhdix: Because every `2^10` is `1024`. And the standard says that `KiB` should be written for "every `1024 bytes` = `1 KB`". It seems to be no longer used everywhere.
(MacOS uses `10^2` and Windows uses `2^10`). But I don't feel good about using `10^2` instead of `2^10`.
https://en.wikipedia.org/wiki/Binary_prefix#Deviation_between_powers_of_1024_and_powers_of_1000

bassosimone: Here's my rule of thumb: either kbit/s (base 10) or KiB/s (base 2). The former is the unit in which they sell you the connection. The latter is the unit in which you download/upload stuff that matters. There is clearly a tension. I've seen tools showing the speed in kbit/s and tools showing it in KiB/s. My personal preference, when writing speed tests, has always been using kbit/s (or Mbit/s, or whatever) under the argument that other speed tests do the same.

xhdix: Okay, so let's consider the `Conventional current`. :P

xhdix:
After a lot of clashes between `vscode` and `git`, it's done. :P

bassosimone:
@xhdix do you know why travis is failing?

experiment/dash/dash_otherwise.go
```diff
@@ -0,0 +1,157 @@
...
+	totalStep      = 15.0
+)
+
+// Config contains the experiment config.
+type Config struct{}
+
+// Simple contains the experiment total summary
+type Simple struct {
+	ConnectLatency  float64 `json:"connect_latency"`
+	MedianBitrate   int64   `json:"cedian_bitrate"`
```
bassosimone: Nit: `median` not `cedian`

experiment/dash/dash_otherwise.go
```diff
@@ -0,0 +1,157 @@
...
+		return client.Error()
+	}
+	data, err := json.Marshal(client.ServerResults())
+	if err != nil {
+		testkeys.Failure = err.Error()
+		return err
+	}
+	sess.Logger.Debugf("Server result: %s", string(data))
+	callbacks.OnProgress(1, "done")
+	testkeys.ClientResolver = sess.Location.ResolverIP
```
bassosimone: This is not needed anymore, we're now using a toplevel key for that

xhdix:
> do you know why travis is failing?
you said: `psiphon should fail because it needs a secret file`

bassosimone:
>> do you know why travis is failing?
> you said: psiphon should fail because it needs a secret file

Yeah, I figured that for myself later :-).

experiment/dash/dash_otherwise.go
xhdix: In the midst of the clashes, they did not change again. :(

xhdix:
```suggestion
	testkeys.ClientResolver = sess.ResolverIP()
```
?
Or delete?
I don't understand: ` we're now using a toplevel key for that`

bassosimone: I meant: delete the line. The code that calls the test is now doing that. Not in the test keys but in the parent object (Measurement). For this reason I said "top level keys", which I can understand why it was not clear at all :)

PR #74, opened 2019-10-02
session: expose the ProbeASN as an integer

bassosimone:
Needed by the probe-cli. I'm trying to reduce the coupling between internals of this repo and the probe-cli, so I am adding a session method to get the ASN as an integer, rather than continuing to assume that the probe-cli can access whatever package in here. I look forward for the cli to be able to use mainly the session and a few other packages possibly.


bassosimone:
@xhdix said this looks good, merging!

PR #92, opened 2019-10-08
Introduce CONTRIBUTING.md and apply its recommendations to dash.go

bassosimone:
After merging #61, I realised there is a bunch of stuff that I didn't properly clarify inside the documentation. Therefore, I sat down and wrote CONTRIBUTING.md.

I have also modified the dash experiment to be in line with CONTRIBUTING.md.

While at it, I also noticed that a contributor who has not access to the psiphon experiment secret would be miserable because tests would always fail. So, make it possible to locally run tests without running the always failing Psiphon test.


bassosimone:
I think @xhdix is the best person to perhaps take a look at that.

PR #94, opened 2019-10-08
Implement DoH experiment

bassosimone:
I'm printing on stdout the stream of events for now. Plan is to
mere this and have machinery. Then, do work encompassing all the
experiments using netx to produce common data format.

Spec for now is very vague with respect to the data format
specifically for this reason. Aim is to revise the spec once
we have written common netx processing code.


bassosimone:
We're not making this experiment as exposed to user, rather it's an automatic follow-up implemented as part of github.com/ooni/netx.

PR #95, opened 2019-10-08
Update to stable m-lab/ndt7-client-go@v0.1.0

bassosimone:
Use the stable client. The spec says to mainly show app-level data. So, I am doing that. Also, the data format has changed with v0.8.0 of the spec.


bassosimone:
@robertodauria can you maybe take a look?

experiment/ndt7/ndt7.go
```diff
@@ -89,13 +89,15 @@ func measure(
...
-		percentage := ev.Elapsed / maxRuntime / 2.0
-		message := fmt.Sprintf(
-			"max-bandwidth (download) %s (RTT min/smoothed/var %.1f/%.1f/%.1f ms)",
-			humanize.SI(float64(ev.BBRInfo.MaxBandwidth), "bit/s"),
-			ev.BBRInfo.MinRTT, ev.TCPInfo.SmoothedRTT, ev.TCPInfo.RTTVar,
-		)
-		callbacks.OnProgress(percentage, message)
+		if ev.AppInfo != nil && ev.Origin == "client" {
+			elapsed := float64(ev.AppInfo.ElapsedTime)/1e06 // to seconds
+			percentage := elapsed/maxRuntime/2.0
```
robertodauria: It's not immediately clear why this percentage is `/2.0`, perhaps a comment would help?

experiment/ndt7/ndt7.go
```diff
@@ -110,12 +112,15 @@ func measure(
...
 		testkeys.Upload = append(testkeys.Upload, ev)
-		percentage := 0.5 + ev.Elapsed/maxRuntime/2.0
-		speed := float64(ev.AppInfo.NumBytes) * 8.0 / ev.Elapsed
-		message := fmt.Sprintf(
-			"upload-speed %s", humanize.SI(float64(speed), "bit/s"),
-		)
-		callbacks.OnProgress(percentage, message)
+		if ev.AppInfo != nil && ev.Origin == "client" {
+			elapsed := float64(ev.AppInfo.ElapsedTime)/1e06 // to seconds
+			percentage := 0.5 + elapsed/maxRuntime/2.0
```
robertodauria: Same here.

bassosimone:
(apparently my branch protection settings have caused the review to be dismissed, which is probably not very useful in this case as too strict.)

PR #97, opened 2019-10-08
Write initial, simple SNI blocking experiment

bassosimone:
Same as other PRs I've opened recently: basic code. We'll wire in
the data-formats when we'll add code to process netx events.

We'll refine and declare stable only with 100% coverage including
making sure we're fully compliant with the spec.

See https://github.com/ooni/spec/pull/159 for the spec.


experiment/sniblocking/sniblocking.go
```diff
@@ -0,0 +1,138 @@
...
+	_, expected := randomErr.(x509.HostnameError)
+	if properErr == nil && expected {
+		tk.Behavior = "normal"
+		return
+	}
+	if properErr != nil && expected {
+		tk.Behavior = "suspicious"
+		return
+	}
+	tk.Behavior = "🤷"
```
hellais: As mentioned in the ooni/spec repo I suggest using simple ascii for this as it will make parsing the data in all languages easier.

bassosimone:
@hellais I've addressed all your feedback, please take another look!

Regarding:

> Beyond that the test looks reasonable to me, I am curious to learn how this is going to be triggered by other tests though as my understanding is that the input to this experiment is some website which is already somehow suspicious.

Yes, this is what I was planning on doing.

bassosimone:
This is implemented as an automatic follow-up experiment in github.com/ooni/netx

PR #99, opened 2019-10-28
Write more abstract and powerful API

bassosimone:
This API allows to manipulate all engine things with very few structs. Also, this API introduces the possibility of querying test options and setting them rather easily.

Also, this API is very tailored to the needs of ooni/probe-cli and anticipates future usage on mobile by trying to expose something that should be easily mockable with go mobile.

### DIFF
--- a/cmd/miniooni/main.go
+++ b/cmd/miniooni/main.go
@@ -2,7 +2,6 @@
 package main
 
 import (
-	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -14,21 +13,8 @@ import (
 
 	"github.com/apex/log"
 
-	"github.com/ooni/probe-engine/experiment"
-	"github.com/ooni/probe-engine/experiment/dash"
-	"github.com/ooni/probe-engine/experiment/example"
-	"github.com/ooni/probe-engine/experiment/fbmessenger"
-	"github.com/ooni/probe-engine/experiment/hhfm"
-	"github.com/ooni/probe-engine/experiment/hirl"
-	"github.com/ooni/probe-engine/experiment/ndt"
-	"github.com/ooni/probe-engine/experiment/ndt7"
-	"github.com/ooni/probe-engine/experiment/psiphon"
-	"github.com/ooni/probe-engine/experiment/telegram"
-	"github.com/ooni/probe-engine/experiment/web_connectivity"
-	"github.com/ooni/probe-engine/experiment/whatsapp"
+	engine "github.com/ooni/probe-engine"
 	"github.com/ooni/probe-engine/httpx/httpx"
-	"github.com/ooni/probe-engine/orchestra/testlists"
-	"github.com/ooni/probe-engine/session"
 
 	"github.com/pborman/getopt/v2"
 )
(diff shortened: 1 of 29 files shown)


cmd/miniooni/main.go
```diff
@@ -191,9 +176,19 @@ func main() {
...
+
+	sess, err := engine.NewSession(engine.SessionConfig{
+		AssetsDir:       tempDir,
+		Logger:          logger,
+		ProxyURL:        proxyURL,
+		SoftwareName:    softwareName,
+		SoftwareVersion: softwareVersion,
+		TempDir:         tempDir,
+		TLSConfig:       tlsConfig,
+	})
```
bassosimone: This is one of the first reasons why I introduced this API: I want to make sure that when we construct a new session we are passing correct parameters and the mandatory ones are there. Also, this setting of options through a struct is more easy to get right than instead passing a long list of options to a constructor.

cmd/miniooni/main.go
```diff
@@ -207,13 +202,13 @@ func main() {
 
 	if !globalOptions.noBouncer {
 		log.Info("Looking up OONI backends")
-		if err := sess.MaybeLookupBackends(ctx); err != nil {
+		if err := sess.MaybeLookupBackends(); err != nil {
```
bassosimone: Since `ooni/probe-cli` is not using any `ctx`, I have hidden the `ctx` for now

cmd/miniooni/main.go
```diff
@@ -222,65 +217,43 @@ func main() {
...
+		log.WithError(err).Fatal("cannot create experiment builder")
+	}
+	if builder.NeedsInput() {
 		if len(globalOptions.inputs) <= 0 {
-			list, err := testlists.NewClient(sess).Do(ctx, sess.ProbeCC(), 100)
+			log.Info("Fetching test lists")
+			config := sess.NewTestListsConfig()
+			config.Limit = 16
+			client := sess.NewTestListsClient()
+			list, err := client.Fetch(config)
```
bassosimone: This is a feature that allows you to query test lists from a session, w/o importing more confusing packages

cmd/miniooni/main.go
```diff
@@ -222,65 +217,43 @@ func main() {
...
-		experiment = ndt.NewExperiment(sess, ndt.Config{})
-	} else if name == "ndt7" {
-		experiment = ndt7.NewExperiment(sess, ndt7.Config{})
-	} else if name == "psiphon" {
-		if _, ok := extraOptions["config_file_path"]; !ok {
-			log.Fatal("psiphon requires the `-O config_file_path=PATH` option")
+	for key, value := range extraOptions {
+		err := builder.SetOptionString(key, value)
+		if err != nil {
+			log.WithError(err).Fatal("cannot set string option")
```
bassosimone: This is the setting of options in action

cmd/miniooni/main.go
```diff
@@ -290,15 +263,15 @@ func main() {
...
+		measurement, err := experiment.Measure(input)
 		if err != nil {
 			log.WithError(err).Warn("measurement failed")
 			continue
 		}
 		measurement.AddAnnotations(annotations)
 		if !globalOptions.noCollector {
 			log.Infof("submitting measurement to OONI collector")
-			if err := experiment.SubmitMeasurement(ctx, &measurement); err != nil {
+			if err := experiment.SubmitAndUpdateMeasurement(measurement); err != nil {
```
bassosimone: New name so it is more clear what it does

experiment.go
```diff
@@ -0,0 +1,402 @@
+// Package engine contains the engine API
+package engine
```
bassosimone: Wrapper for `experiment/experiment.go` with ability of building from a builder

experiment/example/example.go
```diff
@@ -21,7 +22,9 @@ const (
 // This contains all the settings that user can set to modify the behaviour
 // of this experiment.
 type Config struct {
-	SleepTime time.Duration
+	ReturnError bool   `ooni:"Toogle to return a mocked error"`
+	Message     string `ooni:"Message to emit at test completion"`
+	SleepTime   int64  `ooni:"Amount of time to sleep for"`
```
bassosimone: The annotation documents a test's option

Issue #25, opened 2019-06-14
Modify regress tests to also run MK nettests


bassosimone:

bassosimone:
This seems difficult to do and I'd like to focus on rewriting tests in Go instead.

PR #104, opened 2019-11-04
collector: unconditionally write the ReportID

bassosimone:
Issue reported by @FedericoCeratto

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -124,13 +124,14 @@ type updateResponse struct {
 }
 
 // SubmitMeasurement submits a measurement belonging to the report
-// to the OONI collector. On success, we will modify the measurement
-// updating its ReportID field. If the collector supports sending back to
-// us a measurement ID, we also update the m.OOID field with it.
+// to the OONI collector. We will unconditionally modify the measurement
+// with the ReportID it should contain. If the collector supports sending
+// back to us a measurement ID, we also update the m.OOID field with it.
 func (r *Report) SubmitMeasurement(
 	ctx context.Context, m *model.Measurement,
 ) error {
 	var updateResponse updateResponse
+	m.ReportID = r.ID
 	err := (&jsonapi.Client{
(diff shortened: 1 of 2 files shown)


FedericoCeratto:
Any functional test that can be added?

bassosimone:
@FedericoCeratto:

> Any functional test that can be added?

Sure, I can and should do this (see https://github.com/ooni/probe-engine/issues/25). If you don't mind, I'd rather merge this critical fix and cut a release of probe-cli as well, so we have this fix in the desktop app, and then focus on adding more tests.

hellais:
In which cases was the reportID missing from the files? Was that the bug?

I think it's OK to merge this without some checks, but we should at least document in this PR what the problem was (and how to reproduce it) so that we know how to implement a test for it in the future.

bassosimone:
> In which cases was the reportID missing from the files? Was that the bug?

When we run MK tests, the code overwrites the full report. This is convenient because it simplifies life but then it also removes the reportID we had already set in Go code.

Another case where it may break: we have a measurement from a previous run where the upload failed and we're now resubmitting with a different measurement ID.

In both cases, the collector will accept the measurements with empty/wrong measurement ID, and then we'll have a measurement with empty/wrong measurement ID saved on disk.

I believe this fix is fine, because of the second case. The first case may call for changing the way in which MK code integrates with Go code. We may also consider that.

I'd actually start from writing functional tests, as @FedericoCeratto says, and that was actually already part of my immediate post-OTF2019 TODO list.

hellais:
Ok understood, thanks for explaining it.

(I assume in the second case where you say "measurement ID" you mean "report ID" as well)

bassosimone:
> (I assume in the second case where you say "measurement ID" you mean "report ID" as well)

Yes, my bad, I'm a bit 💤 today!

bassosimone:
I'm going to proceed with merging and then I'll follow up with @FedericoCeratto for more tests

Issue #142, opened 2019-12-03
telegram_otherwise_test.go: flaky test


bassosimone:
I've seen this more than once:

```
-- FAIL: TestUnitProcessallWithMixedResults (0.00s)
    telegram_otherwise_test.go:165: invalid TelegramWebFailure
FAIL
```

See:

- https://travis-ci.org/ooni/probe-engine/builds/620071781

- https://travis-ci.org/ooni/probe-engine/builds/620153895#L611

This is what I know for now:

1. cannot reproduce on macOS with `go test -race ./...` when telegram is manually modified to always run the Go version rather than the C++ version

2. does not happen on Linux with `CGO_ENABLED=0 go test -race ./...`

3. seems to happen on travis (four times in ~1 month)

bassosimone:
Did see the failure again. It was like:

```
84.16s$ go test -coverprofile=coverage.cov -coverpkg=./... ./...
ok  	github.com/ooni/probe-engine	29.011s	coverage: 63.7% of statements in ./...
ok  	github.com/ooni/probe-engine/bouncer	0.444s	coverage: 3.4% of statements in ./...
?   	github.com/ooni/probe-engine/cmd/miniooni	[no test files]
ok  	github.com/ooni/probe-engine/collector	0.392s	coverage: 5.7% of statements in ./...
ok  	github.com/ooni/probe-engine/experiment	4.895s	coverage: 30.1% of statements in ./...
ok  	github.com/ooni/probe-engine/experiment/dash	0.008s	coverage: 7.1% of statements in ./...
ok  	github.com/ooni/probe-engine/experiment/example	6.058s	coverage: 28.1% of statements in ./...
ok  	github.com/ooni/probe-engine/experiment/fbmessenger	0.008s	coverage: 1.3% of statements in ./...
ok  	github.com/ooni/probe-engine/experiment/handler	0.007s	coverage: 1.4% of statements in ./...
ok  	github.com/ooni/probe-engine/experiment/hhfm	0.008s	coverage: 1.3% of statements in ./...
ok  	github.com/ooni/probe-engine/experiment/hirl	0.007s	coverage: 1.3% of statements in ./...
?   	github.com/ooni/probe-engine/experiment/httpheader	[no test files]
ok  	github.com/ooni/probe-engine/experiment/mkevent	0.008s	coverage: 3.3% of statements in ./...
[...]

PR #151, opened 2019-12-03
oodatamodel: handle binary header values

### LINKED ISSUE #110: oodatamodel.go: deal with binary header values
Following up from https://github.com/ooni/probe-engine/pull/103#discussion_r347795184

bassosimone:
Unsure whether this could happen, but robustness is usually good.

Closes #110

### DIFF
--- a/experiment/oodatamodel/oodatamodel.go
+++ b/experiment/oodatamodel/oodatamodel.go
@@ -62,17 +62,17 @@ type HTTPTor struct {
 	IsTor    bool    `json:"is_tor"`
 }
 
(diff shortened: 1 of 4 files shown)


bassosimone:
I am going to merge with failing CI tests. The failure is clearly #142 and it seems I need to deal with this failure as an interrupt and make it working again because it's not very frequent.

bassosimone:
Of course, now it's all green :-). Well, at least I've started being prepared for #142 next time that it surfaces and have also bumped psiphon deadline; we'll see if that's enough!

Issue #148, opened 2019-12-03
orchestra: implement authenticated requests


bassosimone:
Part of #109

bassosimone:
The code merged in master (and specifically all the pull requests referenced above) contains enough functionality to allow authenticated requests. A major issue is that we are not storing the credentials on persistent storage, as described by https://github.com/ooni/probe-engine/issues/164. This does not prevent us from implementing an authenticated API for Psiphon and from using it, except that it will create a new ClientID for every session, thus creating pressure on the database. For this reason, I am going to call this issue completed, but I have created #164 as a next-cycle follow-up.

PR #156, opened 2019-12-07
[API CHANGE] Refactor code for fetching test lists

bassosimone:
1. using a "client" for a test lists query is overkill, instead just
have a single function with config and possibly result (the same
concept can be applied to other places in the codebase, in which we
create "client" instances we don't actually need)

2. make sure the toplevel API can be used from mobile code, i.e., avoid
using a vector that gomobile cannot bind

3. verify, by looking at changed tests and miniooni, that the result
API is still reasonably non-painful to use

4. make sure that the internal implementation details are part of the
internal package: we're currently exposing too much internals

These changes are yak shaving I implemented while starting the work
to implement authenticated request for orchestra (i.e. #148)


bassosimone:
I am going to merge with failing Travis because the failed build is useful to investigate #142.

PR #157, opened 2019-12-08
orchestra: implement registering a probe

bassosimone:
Part of #148


bassosimone:
The [failed build](https://travis-ci.org/ooni/probe-engine/builds/622294409#L611) contains a relapse of bug #142, which will be addressed soon. That is the only error we got.

PR #160, opened 2019-12-09
orchestra/testlists/urls consistent w/ other orchestra

bassosimone:
Make sure the code style and structure is consistent among all the
orchestra building blocks by always returning a pointer.


bassosimone:
The [build failure](https://travis-ci.org/ooni/probe-engine/builds/622541163#L628) is because coveralls was not able to correctly find the Travis job, so it's not our fault.

PR #190, opened 2019-12-26
resources: update to ooni/probe-assets@20191226162429

bassosimone:


bassosimone:
The coverage is made worst by the fact that the build initially failed.

PR #201, opened 2020-01-08
Fork ooni/netx/x/{logger,porcelain}

bassosimone:
These are two experimental packages that are too bound to this repo
to be kept into the other repo.

So, for them here. This is yak shaving for #2 because we want to add
functionality to perform TCPConnect like tests.

I didn't want to ping-pong the implementation of that, so I decided
the time was come to have our own porcelain and logger.


internal/netxlogger/netxlogger.go
```diff
@@ -0,0 +1,188 @@
+// Package netxlogger is a logger for netx events.
+//
+// This package is a fork of github.com/ooni/netx/x/logger where
+// we applied ooni/probe-engine specific customisations.
```
bassosimone: This is aspirational. Currently the file is copied verbatim.

internal/oonitemplates/oonitemplates.go
```diff
@@ -0,0 +1,421 @@
+// Package oonitemplates contains templates for experiments.
+//
+// Every experiment should possibly be based on code inside of
+// this package. In the future we should probably unify the code
+// in here with the code in oodatamodel.
+//
+// This has been forked from ooni/netx/x/porcelain because it was
+// causing too much changes to keep this code in there.
```
bassosimone: This is exactly like in netx, except I changed the toplevel comment

Issue #226, opened 2020-01-12
Write tor experiment (episode 1)


bassosimone:
We have discussed with @hellais that, rather than working on separate tests for connecting using TCP to bridges (see #2) and for checking whether obfs4 works (see #90), we'd be better off with writing a single Tor experiment that works as follows:

1. read a list of Tor related targets to check (directory authorities addresses, vanilla bridges, obfs4 bridges, etc.)

2. for each target, check whether there is TCP level connectivity (what #2 did) and then do something else to gain more confidence we can use it (what #90 did for obfs4)

The main rationale of this design is to have a single place experiment that gives a holistic view of how to best configure/use tor in a specific context.

I have been working towards implementing this experiment for a week now, and it seems to me the time has come for replacing #2 and #90 with this issue. I will accordingly strip #2 and #90 of their priority and effort estimations and mark them as wontfix. This issue will be the right place where to measure our progress towards competing this more ambitious goal.

bassosimone:
# Progress tracking

- [ ] placeholder

bassosimone:
The first part of work has been done (see https://github.com/ooni/probe-engine/issues/226#issuecomment-573410230). The follow-up issue for next sprint is https://github.com/ooni/probe-engine/issues/276. (This is an indication very large issues should probably be broken down into smaller issues.)

PR #227, opened 2020-01-12
Implement the new tor experiment

bassosimone:
Closes https://github.com/ooni/probe-engine/issues/226 and https://github.com/ooni/probe-engine/issues/276

### DIFF
--- a/experiment/tor/tor.go
+++ b/experiment/tor/tor.go
@@ -17,6 +17,7 @@ import (
 	"github.com/ooni/probe-engine/internal/netxlogger"
 	"github.com/ooni/probe-engine/internal/oonidatamodel"
 	"github.com/ooni/probe-engine/internal/oonitemplates"
+	"github.com/ooni/probe-engine/internal/orchestra"
 	"github.com/ooni/probe-engine/model"
 	"github.com/ooni/probe-engine/session"
 )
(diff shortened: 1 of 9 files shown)


internal/orchestra/testlists/tor/tor.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/ooni/probe-engine/model"
+)
+
+// Config contains settings.
+type Config struct{}
+
+var staticTestingTargets = []model.TorTarget{
+	// TODO(bassosimone): this is a public working bridge we have found
+	// with @hellais. We should ask @phw whether there is some obfs4 bridge
+	// dedicated to integration testing that we should use instead.
```
hellais: @nullhypothesis do you have tips for what better bridge we could use in here?

experiment/tor/tor.go
```diff
@@ -0,0 +1,241 @@
...
+			Method:                  "GET",
+			URL:                     url.String(),
+			UserAgent:               httpheader.RandomUserAgent(),
+		})
+		tk, err = r.TestKeys, r.Error
+	case "or_port":
+		r := oonitemplates.TLSConnect(ctx, oonitemplates.TLSConnectConfig{
+			Address:            target.Address,
+			Beginning:          rc.measurement.MeasurementStartTimeSaved,
+			InsecureSkipVerify: true, // TODO(bassosimone): can we avoid it?
```
hellais: I think we probably still want to detect if there is a MITM and in that case still record the certificates.

That said maybe we can do some validation logic which is closer to the real validation logic that tor does. I don't actually know the details of this, but perhaps @nullhypothesis can help us figure this out.

experiment/tor/tor.go
```diff
@@ -0,0 +1,241 @@
...
+	measurement *model.Measurement,
+	callbacks handler.Callbacks,
+	targets []model.TorTarget,
+) error {
+	// run measurements in parallel
+	var waitgroup sync.WaitGroup
+	rc := newResultsCollector(sess, measurement, callbacks)
+	waitgroup.Add(len(targets))
+	workch := make(chan model.TorTarget)
+	const parallelism = 2
```
hellais: Perhaps we should define the parallelism constant higher up inside of this file with a comment above it saying that we will be measuring two targets at the same time?

experiment/tor/tor.go
bassosimone: So, the certificate is already included into the report. Yes, we can try to mimic the logic that tor would use, and we can event consider doing this at a later stage.

experiment/tor/tor.go
bassosimone: Sure

bassosimone: Sure!

internal/orchestra/testlists/tor/tor.go
NullHypothesis: The most reliable bridges are our default bridges because we monitor them and poke their operators as soon as one goes offline:
https://trac.torproject.org/projects/tor/wiki/doc/TorBrowser/DefaultBridges
You may want to pick David Goulet's bridge `CDF2E852BF539B82BD10E27E9115A31734E378C2`. It has been online for a long time, it's reliable, and it's easy to get in touch with David.

experiment/tor/tor.go
NullHypothesis: It would be very useful to know if a probe's TLS handshake got MitM-ed. Regarding tor's validation logic: to verify if a given relay's certificate is authentic, we must fetch the relay's server descriptor, [take its ed25519 signing key](https://gitweb.torproject.org/torspec.git/tree/cert-spec.txt#n136) and make sure that the TLS certificate is signed by this signing key. As long as we have the certificate and its signature, we should be able to verify after the fact if it's authentic.

experiment/tor/tor.go
bassosimone: done!

experiment/tor/tor.go
bassosimone: Okay so we have the certificate and its signature!

experiment/tor/tor_test.go
```diff
@@ -0,0 +1,384 @@
...
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+var staticTestingTargets = []model.TorTarget{
+	// TODO(bassosimone): this is a public working bridge we have found
+	// with @hellais. We should ask @phw whether there is some obfs4 bridge
+	// dedicated to integration testing that we should use instead.
```
bassosimone: Here we should use CDF2E852BF539B82BD10E27E9115A31734E378C2 instead

internal/orchestra/testlists/tor/tor.go
bassosimone: done!

experiment/tor/tor_test.go
bassosimone: done!

PR #295, opened 2020-01-27
tor: add support for or_port_dirauth

bassosimone:
See https://github.com/ooni/sysadmin/commit/bf082ff0953d72bdf67773b2527716e50646908e

Part of https://github.com/ooni/probe/issues/973

### DIFF
--- a/experiment/tor/tor.go
+++ b/experiment/tor/tor.go
@@ -44,6 +44,7 @@ type TargetResults struct {
 	Queries        oonidatamodel.DNSQueriesList    `json:"queries"`
 	Requests       oonidatamodel.RequestList       `json:"requests"`
 	TargetAddress  string                          `json:"target_address"`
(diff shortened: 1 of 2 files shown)


experiment/tor/tor.go
```diff
@@ -230,7 +239,7 @@ func (rc *resultsCollector) defaultFlexibleConnect(
 			Path:   "/tor/status-vote/current/consensus.z",
 			Scheme: "http",
 		}
-		const snapshotsize = 1 << 16 // no need to include all in report
+		const snapshotsize = 1 << 8 // no need to include all in report
```
bassosimone: I chose to reduce the snapshot as we've discussed with @hellais that we wanted to compute the simhash (which has been scheduled as part of the next cycle) and since we may not be able to deflate anyway

Issue #267, opened 2020-01-14
measurementkit: build in ordinary linux systems


ja-pa:
Hi, I'm trying to create a package of probe-cli for OpenWrt. During compilation, probe-engine was failing on library linking  because of hardcoded path definition in https://github.com/ooni/probe-engine/blob/master/measurementkit/task_cgo.go
After adding these cgo definitions I was able to compile and run test probe-engine on OpenWrt.

```
@@ -38,6 +38,18 @@ import (
 	// #cgo linux,amd64 LDFLAGS: /usr/lib/libevent_extra.a
 	// #cgo linux,amd64 LDFLAGS: /usr/lib/libevent_pthreads.a
 	// #cgo linux,amd64 LDFLAGS: /lib/libz.a
+	//
+	// #cgo LDFLAGS: -lmeasurement_kit
+	// #cgo LDFLAGS: -lmaxminddb
+	// #cgo LDFLAGS: -lcurl
+	// #cgo LDFLAGS: -lnghttp2
+	// #cgo LDFLAGS: -levent_openssl
+	// #cgo LDFLAGS: -lssl
+	// #cgo LDFLAGS: -lcrypto
+	// #cgo LDFLAGS: -levent_core
+	// #cgo LDFLAGS: -levent_extra
+	// #cgo LDFLAGS: -levent_pthreads
+	// #cgo LDFLAGS: -lz
 	"C"
 	"errors"
 	"unsafe"
```
Is it possible to integrate similar changes to probe-engine master?

bassosimone:
We would need to test whether this breaks out builds or not. The reasons paths for Linux are hard code is because we want to create a statically linked binary. I assume the build works for you if you export such LDFLAGS into an environment variable and run `go build`. Is that correct?

ja-pa:
Thanks for the explanation. Yes, build works +- as you described. I'm thinking if it would be possible to solve this by some specific tag? But I don't know golang too much. . .

bassosimone:
I will take a look

bassosimone:
Should be fixed in https://github.com/ooni/probe-engine/pull/333 and https://github.com/ooni/probe-cli/pull/112. Awaiting confirmation from @ja-pa that it's WAI on OpenWrt.

PR #333, opened 2020-02-14
task_cgo.go: allow compiling on ordinary Linux systems

bassosimone:
Part of #267


bassosimone:
@ja-pa can you check whether this diff works for you? I figured that in the most general Linux case there are dynamic libraries, so it should be enough to `-lmeasurement_kit`.

I have confirmed that this works for us if we change our build script to add `-tags ooni`. The default now is to serve the needs of an user compiling on Linux with MK installed.

ja-pa:
Hi, I tried that and it works ok. There is still panic with running test on OpenWrt but I think that it's a separate issue (https://github.com/openwrt/packages/pull/11116#discussion_r378588986)

bassosimone:
@ja-pa thanks!

PR #364, opened 2020-02-22
oonimkall: be liberal in the settings we accept

bassosimone:
We were too strict and this means coding the Android app to
support MK and probe-engine at the same time is painful.

This is part of https://github.com/measurement-kit/measurement-kit/issues/1913.

### DIFF
--- a/oonimkall/chanlogger.go
+++ b/oonimkall/chanlogger.go
@@ -11,7 +11,6 @@ type chanLogger struct {
 	hasinfo    bool
 	haswarning bool
 	out        chan<- *eventRecord
-	settings   *settingsRecord
 }
 
 // Debug implements Logger.Debug
@@ -67,15 +66,14 @@ func (cl *chanLogger) Warnf(format string, v ...interface{}) {
 
 // newChanLogger creates a new ChanLogger instance.
 func newChanLogger(
-	emitter *eventEmitter, settings *settingsRecord,
+	emitter *eventEmitter, logLevel string,
 	out chan<- *eventRecord,
 ) *chanLogger {
 	cl := &chanLogger{
 		emitter:  emitter,
 		out:      out,
-		settings: settings,
 	}
-	switch settings.LogLevel {
+	switch logLevel {
 	case "DEBUG", "DEBUG2":
 		cl.hasdebug = true
 		fallthrough
--- a/oonimkall/runner.go
+++ b/oonimkall/runner.go
@@ -47,7 +47,7 @@ func newRunner(settings *settingsRecord, out chan<- *eventRecord) *runner {
 	}
 }
 
-func (r *runner) hasUnsupportedSettings() (unsupported bool) {
+func (r *runner) hasUnsupportedSettings(logger *chanLogger) (unsupported bool) {
 	sadly := func(why string) {
 		r.emitter.EmitFailureStartup(why)
 		unsupported = true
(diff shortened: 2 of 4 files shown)


oonimkall/chanlogger.go
```diff
@@ -11,7 +11,6 @@ type chanLogger struct {
 	hasinfo    bool
 	haswarning bool
 	out        chan<- *eventRecord
-	settings   *settingsRecord
```
bassosimone: We were actually not using this field

oonimkall/runner.go
```diff
@@ -47,7 +47,7 @@ func newRunner(settings *settingsRecord, out chan<- *eventRecord) *runner {
 	}
 }
 
-func (r *runner) hasUnsupportedSettings() (unsupported bool) {
+func (r *runner) hasUnsupportedSettings(logger *chanLogger) (unsupported bool) {
```
bassosimone: A bunch of cases have become warnings because otherwise it's complex to keep MK/non-MK code together. We may want to consider removing these fields in the future, maybe using something like an API number to kick out old clients.

oonimkall/runner.go
```diff
@@ -59,34 +59,31 @@ func (r *runner) hasUnsupportedSettings() (unsupported bool) {
...
+		logger.Warn("Options.ProbeNetworkName: not supported")
 	}
-	if r.settings.Options.RandomizeInput != nil {
+	if r.settings.Options.RandomizeInput != false {
 		sadly("Options.RandomizeInput: not supported")
 	}
-	if r.settings.OutputFilePath != "" {
-		sadly("OutputFilePath: not supported")
+	if r.settings.OutputFilepath != "" && r.settings.Options.NoFileReport == false {
+		sadly("OutputFilepath && !NoFileReport: not supported")
```
bassosimone: These two cases have been conflated together because this seems to me the best way to guarantee a backward compatible semantics. The OONI app is setting `NoFileReport` to `true`. Historically MK has used an empty output file name to mean that MK would choose a name (what a terrible choice... 🤦‍♀🤦‍♀🤦‍♀🤦‍♀🤦‍♀🤦‍♀🤦‍♀).

The only aspect that makes me a little sorry here is that `NoFileReport` defaults to `false`, so a user integrating this library needs to set this field explicitly to `true` for the test to work. Anyway, it probably makes sense in the future to support the `OutputFilepath` option by default, but it mainly does not make sense to work on this now, since the OONI app manages the file.

As an historical note, the reason why we have chosen at a later stage to avoid writing on disk was because MK was supposed being an engine, therefore the app should have been responsible of the file system. However, it's not always that easy, and both probe-engine and MK manage a bunch of files, so maybe writing the result makes sense.

What we can actually do is to change the name of the field like `output_file_path` and give it the correct semantics, i.e., that we are going to write the file if the string is not empty.

oonimkall/runner.go
```diff
@@ -114,13 +111,13 @@ func (r *runner) Run(ctx context.Context) {
 	// TODO(bassosimone): honour context
 	// TODO(bassosimone): intercept all options we ignore
 
+	logger := newChanLogger(r.emitter, r.settings.LogLevel, r.out)
```
bassosimone: I have changed `newChanLogger`'s second argument to take the log level directly because

1. the default is WARNING when the log level is not recognised

2. otherwise it felt very weird that the we were passing to a constructor the settings before "validating" them in hasUnsupportedSettings (the name of the method actually is slightly different than validate settings, hence the quotes, and the previous code could probably have been fine anyway, but I really wanted to remove ambiguity.)

oonimkall/runner_test.go
```diff
@@ -4,49 +4,60 @@ import (
...
 			ProbeIP:          "127.0.0.1",
 			ProbeNetworkName: "XXX",
-			RandomizeInput:   &randomizeInput,
+			RandomizeInput:   true,
 		},
-		OutputFilePath: "foo",
+		OutputFilepath: "foo",
 	}
-	numseen := make(chan int)
 	go func() {
```
bassosimone: The implementation of this test was unnecessarily complex and ineffective. Now it should be more strict and easier.

bassosimone:
This is working, so merging!

Issue #399, opened 2020-03-10
probe-engine crash with x86 simulator


bassosimone:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xbe9d96bc]
goroutine 5 [running]:
runtime/internal/atomic.Xadd64(0x8a9885f4, 0x1, 0x0, 0xbe9e31e7, 0x8a80e310)
	/Users/runner/hostedtoolcache/go/1.14.0/x64/src/runtime/internal/atomic/asm_386.s:105 +0xc
github.com/ooni/probe-engine.(*Session).queryBouncer(0x8a9885b0, 0xbf54ac60, 0x8a8a6010, 0x8a842cb0, 0x8a85a380, 0x0)
	/Users/runner/runners/2.165.2/work/probe-engine/probe-engine/session.go:485 +0x45
github.com/ooni/probe-engine.(*Session).maybeLookupCollectors(0x8a9885b0, 0xbf54ac60, 0x8a8a6010, 0x0, 0xbf23bc65)
	/Users/runner/runners/2.165.2/work/probe-engine/probe-engine/session.go:425 +0x94
github.com/ooni/probe-engine.(*Session).maybeLookupBackends(0x8a9885b0, 0xbf54ac60, 0x8a8a6010, 0xbf481a80, 0x8a80e300)
	/Users/runner/runners/2.165.2/work/probe-engine/probe-engine/session.go:413 +0x32
github.com/ooni/probe-engine.(*Session).MaybeLookupBackends(...)
	/Users/runner/runners/2.165.2/work/probe-engine/probe-engine/session.go:198
github.com/ooni/probe-engine/oonimkall.(*runner).Run(0x8a842fc4, 0xbf54ac40, 0x8a8581e0)
	/Users/runner/runners/2.165.2/work/probe-engine/probe-engine/oonimkall/runner.go:146 +0x1805
github.com/ooni/probe-engine/oonimkall.StartTask.func1(0x8a858200, 0x8a82a590, 0x8a878000, 0xbf54ac40, 0x8a8581e0)
	/Users/runner/runners/2.165.2/work/probe-engine/probe-engine/oonimkall/task.go:48 +0x120
created by github.com/ooni/probe-engine/oonimkall.StartTask
	/Users/runner/runners/2.165.2/work/probe-engine/probe-engine/oonimkall/task.go:44 +0x199
[...]

PR #414, opened 2020-03-18
Replace the atomic package with atomicx

bassosimone:
The atomicx package is less efficient however it completely avoids
using atomic semantics on int64, which may lead to crash if the
variable we are accessing is not properly aligned.

See https://golang.org/pkg/sync/atomic/#pkg-note-BUG, which reads:

> On x86-32, the 64-bit functions use instructions unavailable
> before the Pentium MMX.
>
> On non-Linux ARM, the 64-bit functions use instructions
> unavailable before the ARMv6k core.
>
> On ARM, x86-32, and 32-bit MIPS, it is the caller's responsibility
> to arrange for 64-bit alignment of 64-bit words accessed atomically. The
> first word in a variable or in an allocated struct, array, or slice can
> be relied upon to be 64-bit aligned.

By using a separate package we are able to possibly even change the
way in which we implement the code, and surely we are able to shield
ourselves from this pitfalls by using a mutex. While in general it
may not be okay to always use a mutex, we mostly use atomic semantics
for counting. That is, we are not using it in performance critical
parts of the tree, hence this fix seems fine to me.

Related issue #399.

### DIFF
--- a/atomicx/atomicx.go
+++ b/atomicx/atomicx.go
@@ -1,0 +1,39 @@
+// Package atomicx contains atomic extensions. The main reason for rolling out
+// this package is to avoid potential crashes when using 32 bit devices where we
+// are atomically accessing a 64 bit variable that is not aligned. The solution
+// to this issue is rather crude: use a normal variable and protect it using a
+// normal mutex. While this could be disappointing in general, it seems fine to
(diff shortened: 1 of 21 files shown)


bassosimone:
We have tested this on x86 simulators with @lorenzoPrimi and it seems to be WAI.

Issue #404, opened 2020-03-13
Record website accessibility performance metrics when using Psiphon


hellais:
This involves configuring a website reachability test to use Psiphon to do the measurement and record that data inside of a Psiphon test result.

We should evaluate if we can provision the sites that are used as targets dynamically.

bassosimone:
When working on redesigning netx, I have payed special attention to make sure this kind of measurements are possibile. The PoC for such redesign is https://github.com/ooni/probe-engine/pull/421.

bassosimone:
This work is mostly done. With the implementation of the `urlgetter` experiment in Sprint 12, we can fetch a measurement over Psiphon and see performance metrics. The next cleanup to come is to reimplement the `Psiphon` experiment as a call to `urlgetter` plus bookkeeping.

bassosimone:
I am adding this issue to sprint 13. Issue https://github.com/ooni/probe-engine/issues/576 will include enhancements such as the Psiphon rewrite that will cause this to happen naturally. As such, we can actually remove any effort to this issue. Effort has been accounted for in other issues.

Issue #359, opened 2020-02-21
PLAN: netx refactoring and improvements


bassosimone:
This issue contains a plan to radically simplify and make netx code more correct (for context the most up-to date design document for the original netx [is available at /legacy/netx/DESIGN.md](https://github.com/ooni/probe-engine/blob/3ab9873adf4bf4ea182597538f9f122a10006caa/legacy/netx/DESIGN.md)).

The companion draft pull request https://github.com/ooni/probe-engine/pull/421 illustrates the principles described here in a PoC implementation. This plan was drafted in the March, 27 week _but it probably still needs some hammering and thinking_. The initial motivation for working on this plan was to solve https://github.com/ooni/probe-engine/issues/125, which proved to be more difficult than expected and which triggered design thinking about netx.

The plan has been implemented since then and https://github.com/ooni/probe-engine/issues/745 is the last issue describing this whole effort. In most cases (I hope) issues related to this plan have been linked to this Epic.

- - -

# context.Context

We should make sure [we use the context as little as possible](https://github.com/ooni/probe-engine/pull/420/files#r399257800). The current implementation is setting a `MeasurementRoot` to the context, but [it is possible to perform the same measurements without](https://github.com/ooni/probe-engine/pull/421/files#r399273153). So, we should not be using the context, because that leads to better code.

# DialTLSContext

Since Go 1.14, the standard library has support for `net·http.Transport.DialTLSContext`, which means that we can use the same TLS dialing code everywhere. Because of the above point of reducing the usage of the Context, we should also actually see to collect as much data as possible without using `httptrace`. Now that we can manage our TLS dial, `httptrace` is certainly useful to see some specific events but luckily is not anymore a hard requirement for observing TLS. See how this is implemented in https://github.com/ooni/probe-engine/pull/421/commits/ac9a8cb266c71ec15e26cf13d3ff3c3d74a27de9.

# Every measurement uses a Transport

We're already doing that with netx. This has the profound implication that we can radically simplify the code. We could have a very simple transport, dialer, etc that just concerns itself with doing its job and logging. Then we can construct much more complex objects when measuring.

```Go
func Get(ctx context.Context, url string, logger Logger) (Results, error) {
  var dialer Dialer = new(net.Dialer)
  dialer = &ErrWrappingDialer{Dialer: dialer}
  dialSaver := &DialSaver{Dialer: dialer}
  dialer = &LogDialer{Dialer: dialSaver, Logger: logger}
[...]

Issue #125, opened 2019-11-28
Uniform bytes measurements and accounting


bassosimone:
In #118 we're measuring the bytes consumed by Telegram, but we're not measuring the bytes consumed by contacting OONI services. We should actually do that, because the original MK implementation was doing that, thus we need to have feature parity.

bassosimone:
Before we can do this now, we need to finish converting the code to `netx`. Moving to Sprint 7.

bassosimone:
Done in https://github.com/ooni/probe-engine/pull/418

bassosimone:
We also need to do some work in probe-cli, reopening

bassosimone:
I realized, in particular, that we're not correctly measuring the bytes in probe-cli. We're missing in particular the bytes used for submitting the report. As discussed with @hellais in chat, we don't want to include in this number of bytes interactions that don't pertain the experiment itself (e.g. downloading the assets). We're going to open an issue to track whether we want to report to the user the total amount of bytes consumed by using OONI (including downloading assets).

bassosimone:
Here's the issue for tracking counting _all_ the bytes consumed: https://github.com/ooni/probe/issues/1069

bassosimone:
This issue ended up being much more complicated than intended. It turns out the design of netx was (1) already rather complicated and (2) did not allow easily to _independently_ count bytes for the experiment and for the session. Because I recognized the importance of reducing the complexity of netx due to both (1) and (2), I spent time designing how to simplify netx rather than working on this issue in this sprint. As part of that, it seems I was able to improve the design significantly (https://github.com/ooni/probe-engine/issues/359) and also to solve the problem at hand (https://github.com/ooni/probe-engine/pull/421/commits/5dbf679ba0b0798148d06c7a9c52837042a7c3dc).

bassosimone:
The detailed plan for implementing this feature in Sprint 10 is the following:

- [ ] import https://github.com/ooni/probe-engine/commit/5dbf679ba0b0798148d06c7a9c52837042a7c3dc as `netx/bytecounter`

- [ ] modify the above diff such that it has its own definition of `Dialer`, so that we don't need to depend on other packages

- [ ] modify the way in which we construct a dialer in `netx/internal/dialer` such that this bytecounting dialer is always applied when we construct the dialer, so we have this functionality

- [ ] move `netx/internal/dialer` to `netx/dialer` and consolidate all dialing functionality together including merging `netx/bytecounter` into this new package

[...]

PR #420, opened 2020-03-24
Proposals for improving netx API and implementation

bassosimone:
This pull request contains a bunch of patches aimed at improving the netx API and implementation. It is meant as a companion of https://github.com/ooni/probe-engine/issues/359, where we specify how such changes should look like. It has been originally triggered by work on https://github.com/ooni/probe-engine/issues/125 and on https://github.com/ooni/probe-engine/issues/404, where I learned a lot about ways in which we can evolve netx. At the moment of opening this PR, the plan is to keep this as a Draft PR for documentation purposes, and cherry-pick specific patches if/when needed.


netx/bytecounting/bytecounting.go
```diff
@@ -0,0 +1,56 @@
...
+
+// DialContext implements measurable.Connector.DialContext
+func (c *Counter) DialContext(
+	ctx context.Context, network, address string) (net.Conn, error) {
+	conn, err := c.Operations.DialContext(ctx, network, address)
+	if err != nil {
+		return nil, err
+	}
+	return &connWrapper{Conn: conn, Counter: c}, err
+}
```
bassosimone: The specific problem of this implementation is that the byte counter is also a dialer, therefore we cannot separate the byte counting and the dialing. However, we sometimes want to account for session and sometimes for both session and experiment. Therefore, this implementation here is not actually meeting our objectives.

netx/errwrapping/errwrapping.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Failure returns the OONI failure string. In particular:
+//
+// 1. if there is no error, it returns ""
+// 2. if the error is not a wrapped error, it returns unknown_failure
+// 3. otherwise, it returns the OONI failure string
+//
+// See the netx/DESIGN.md for more info.
+func Failure(err error) string {
```
bassosimone: `Failure` and `Operation` are quite useful and there should probably be an issue about them.

netx/eventsaving/eventsaving.go
```diff
@@ -0,0 +1,78 @@
...
+	"time"
+
+	"github.com/ooni/probe-engine/netx/measurable"
+)
+
+// An Event describes something that occurred.
+type Event struct {
+	Op   string
+	Time time.Time
+}
```
bassosimone: It is probably more convenient to be able to wrap every Dialer independently of, say, a Resolver, because this gives us more flexibility. Also, a best practice of Go is to keep interfaces small, so...

netx/measurable/measurable.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+// Operations contains measurable operations
+type Operations interface {
+	LookupHost(ctx context.Context, domain string) ([]string, error)
+	DialContext(ctx context.Context, network, address string) (net.Conn, error)
+	Handshake(ctx context.Context, conn net.Conn, config *tls.Config) (
+		net.Conn, tls.ConnectionState, error)
+	RoundTrip(req *http.Request) (*http.Response, error)
+}
```
bassosimone: As said earlier it would be more flexible to have several smaller interfaces.

netx/measurable/measurable.go
```diff
@@ -0,0 +1,219 @@
...
+// WithOperations returns a new context using the specified measurable operations.
+func WithOperations(ctx context.Context, operations Operations) context.Context {
+	return context.WithValue(ctx, operationsKey{}, operations)
+}
+
+// ContextOperations returns the Operations associated with the context.
+func ContextOperations(ctx context.Context) Operations {
+	operations, _ := ctx.Value(operationsKey{}).(Operations)
+	return operations
+}
```
bassosimone: This design allows us to change basically everything with every request using a context. This is nice but at the same time it becomes very confusing to understand what is happening when you are reading the code. It seems that we should not pursue this strategy and rather see to use the context as sparingly as possible.

bassosimone:
There is too much magic when we're using the context to possibly hijack every operation. It seems better to use the context much less frequently as we do in https://github.com/ooni/probe-engine/pull/421

PR #421, opened 2020-03-26
PoC: netx rewrite with better design principles

bassosimone:
This is the companion PR and PoC of https://github.com/ooni/probe-engine/issues/359. Ideally, this PR won't be merged but will remain as a PoC. Individual commits may be cherry-picked into other feature branches though.


internal/httptransport/httptransport.go
```diff
@@ -0,0 +1,250 @@
...
+		},
+		WroteRequest: func(httptrace.WroteRequestInfo) {
+			mu.Lock()
+			rte.WroteRequestTime = time.Now()
+			mu.Unlock()
+		},
+	}
+	rte.URL = req.URL.String()
+	rte.RoundTripStartTime = time.Now()
+	req = req.WithContext(httptrace.WithClientTrace(req.Context(), tracer))
```
bassosimone: This creates a new context. While the callbacks are merged, this context only applies to this request. This is nice because in principle a user could choose to set their own callbacks (which in practice won't happen I think).

internal/measurer/measurer.go
```diff
@@ -0,0 +1,127 @@
...
+	Logger Logger
+}
+
+// Results contains the measurement results.
+type Results struct {
+	Connects       []dialer.Events
+	HTTPBodies     []httptransport.BodySnapshot
+	HTTPRoundTrips []httptransport.Events
+	Resolutions    []resolver.Events
+	TLSHandshakes  []tlsdialer.Events
```
bassosimone: This is unsurprisingly similar to what we have in oonitemplates.

internal/measurer/measurer.go
```diff
@@ -0,0 +1,127 @@
...
+	g.newHTTPClient(d, td)
+	return g
+}
+
+func (g *getter) newResolver() resolver.Resolver {
+	var r resolver.Resolver = resolver.Base()
+	r = resolver.ErrWrapper{Resolver: r}
+	g.resolverSaver = &resolver.EventsSaver{Resolver: r}
+	r = g.resolverSaver
+	r = resolver.LoggingResolver{Resolver: r, Logger: g.config.Logger}
```
bassosimone: Here of course we can add fallbacks

internal/measurer/measurer.go
```diff
@@ -0,0 +1,127 @@
...
+	h = g.tlsDialerSaver
+	h = tlsdialer.LoggingHandshaker{Handshaker: h, Logger: g.config.Logger}
+	return tlsdialer.StdlibDialer{
+		CleartextDialer: d,
+		Handshaker:      h,
+	}
+}
+
+func (g *getter) newHTTPClient(d dialer.Dialer, td tlsdialer.Dialer) {
+	var txp httptransport.Transport = httptransport.NewBase(d, td)
```
bassosimone: Here we should probably disable the Proxy

internal/measurer/measurer.go
```diff
@@ -0,0 +1,127 @@
...
+	txp = httptransport.ErrWrapper{Transport: txp}
+	g.httpTransportSaver = &httptransport.EventsSaver{Transport: txp}
+	txp = g.httpTransportSaver
+	txp = httptransport.HeaderAdder{Transport: txp, UserAgent: "miniooni/0.1.0-dev"}
```
bassosimone: Here we should choose the proper header instead

internal/tlsdialer/tlsdialer.go
```diff
@@ -0,0 +1,199 @@
...
+type StdlibDialer struct {
+	CleartextDialer CleartextDialer
+	Handshaker      Handshaker
+}
+
+// DialContext implements Dialer.DialContext
+func (d StdlibDialer) DialContext(
+	ctx context.Context, network, address string) (net.Conn, error) {
+	return d.DialTLSContext(ctx, network, address)
+}
```
bassosimone: This is not needed here

internal/tlsdialer/tlsdialer.go
```diff
@@ -0,0 +1,199 @@
...
+// DialTLSContext implements TLSDialer.DialTLSContext.
+func (d StdlibDialer) DialTLSContext(
+	ctx context.Context, network, address string) (net.Conn, error) {
+	return d.DialTLSContextWithConfig(ctx, network, address, nil)
+}
+
+// DialTLSContextWithConfig dials a TLS connection using the specified config, which
+// may be nil, in which case we will use a default config. If config.NextProtos is not
+// specified, then we'll advertise as next protocols "h2" and "http/1.1".
+func (d StdlibDialer) DialTLSContextWithConfig(
```
bassosimone: Should not be public, we can change the config on the fly with the handshaker, right?

internal/tlsdialer/tlsdialer.go
```diff
@@ -0,0 +1,199 @@
...
+	ctx context.Context, network, address string, config *tls.Config) (net.Conn, error) {
+	hostname, _, err := net.SplitHostPort(address)
+	if err != nil {
+		return nil, err
+	}
+	if config == nil {
+		config = new(tls.Config)
+	}
+	config = config.Clone()
```
bassosimone: This should be else

internal/measurer/measurer.go
```diff
@@ -0,0 +1,127 @@
...
+	g.httpBodySaver = &httptransport.SnapshotSaver{Transport: txp}
+	txp = g.httpBodySaver
+	txp = httptransport.Logging{Transport: txp, Logger: g.config.Logger}
+	g.httpClient = &http.Client{Transport: txp}
```
bassosimone: This client should also include a properly configured cookie jar

internal/dialer/dialer.go
```diff
@@ -0,0 +1,256 @@
...
+	Network       string
+	StartTime     time.Time
+}
+
+// Service is the dialing service. This is useful to create a flexible HTTP
+// transport, for which the dialer is undertermined. Just pass an instance of
+// this struct to httptransport.NewBase(). Every dial request will block and
+// possibly timeout, until a single goroutine takes ownership of this instance
+// of Service by running service.Loop. Because service.Loop takes in input
+// a dialer, it means you can customize dial depending on circumstances.
```
bassosimone: This service should help to rather easily (and explicitly) solve https://github.com/ooni/probe-engine/issues/125

internal/dialer/dialer.go
```diff
@@ -0,0 +1,256 @@
...
+			} else {
+				select {
+				case req.conn <- conn:
+				default:
+					conn.Close() // cannot sent? we cleanup
+				}
+			}
+		}
+	}
+}
```
bassosimone: Alternative, perhaps better implementation:

```Go
func (d Service) Start(ctx context.Context, dialer Dialer) error {
	select {
	case d.sync <- true:
	default:
		return ErrBusy
	}
	go func() {
		defer func() { <-d.sync }()
		for {
			select {
			case <-ctx.Done():
				return nil // okay to starve requests because they have ctx
			case req := <-d.ch:
				conn, err := dialer.DialContext(ctx, req.network, req.address)
				if err != nil {
					select {
					case req.err <- err:
					default:
					}
				} else {
					select {
					case req.conn <- conn:
					default:
						conn.Close() // cannot sent? we cleanup
					}
				}
			}
		}
	}()
	return nil
}
```

internal/httpclient/httpclient.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/ooni/probe-engine/internal/tlsdialer"
+)
+
+// Logger is the logger interface assumed by this package
+type Logger interface {
+	Debugf(format string, v ...interface{})
+}
+
+// New creates a new HTTP client.
+func New(logger Logger) *http.Client {
```
bassosimone: This should probably take in input a `Dialer`, so we can create a client with undetermined real `Dialer` by using the `Service` and we can hence control a bunch of stuff that happens, as well as wrap the `Conn` before passing it back to the HTTP code.

internal/dialer/dialer.go
bassosimone: So:

```Go
func (d Service) Start(ctx context.Context, dialer Dialer) error {
	select {
	case d.sync <- true:
	default:
		return ErrBusy
	}
	go func() {
		defer func() { <-d.sync }()
		d.loop(ctx, dialer)
	}()
	return nil
}

func (d Service) loop(ctx context.Context, dialer Dialer) {
	var wg sync.WaitGroup
	for {
		select {
		case <-ctx.Done():
			wg.Wait()
			return nil // okay to starve requests because they have ctx
		case req := <-d.ch:
			wg.Add(1)
			go func() {
				defer wg.Done()
				d.dial(ctx, dialer, req, &wg)
			}()
		}
	}
}

func (d Service) dial(
	ctx context.Context, dialer Dialer, req serviceRequest, wg *sync.WaitGroup) {
	conn, err := dialer.DialContext(ctx, req.network, req.address)
	if err != nil {
		select {
		case req.err <- err:
		default:
		}
	} else {
		select {
		case req.conn <- conn:
		default:
			conn.Close() // cannot sent? we cleanup
		}
	}
}
```

internal/dialer/dialer.go
bassosimone: However, it seems that the solution using a bytecounter is more idiomatic Go and easy to reason about than this one, albeit I admit I quite like the idea of returning a dialer using a channel :^)

internal/measurer/measurer.go
```diff
@@ -0,0 +1,127 @@
...
+	HTTPBodies     []httptransport.BodySnapshot
+	HTTPRoundTrips []httptransport.Events
+	Resolutions    []resolver.Events
+	TLSHandshakes  []tlsdialer.Events
+}
+
+// Get performs a GET request and returns the measurement results. The error
+// signals whether the request failed. You will get results in any case, since
+// the purpose of this function is to measure anomalies :^).
+func Get(url string, config Config) (Results, error) {
```
bassosimone: This function shows that we can obtain the same measurements we're currently obtaining with the `MeasurementRoot` stored inside the context without using the context at all (at least at this level of abstraction).

bassosimone:
This PoC has been superseded by work done in the context of https://github.com/ooni/probe-engine/issues/509

PR #425, opened 2020-03-30
Update README.md

darvari128:
(just for test)


bassosimone:
Thanks! So it seems the builds w/ secrets don't even start!

PR #426, opened 2020-03-30
Remove an empty line (this is a test, please ignore it)

robertodauria:


bassosimone:
Thanks! So it seems the builds w/ secrets don't even start!

PR #452, opened 2020-04-02
Experiments in Go now use contex to report bytes usage

bassosimone:
Luckily the mechanism for registering a byte counter with the context is
idempotent, so it does not matter a great deal if by mistake we, say, register
twice the experiment. In fact, there is one slot for the experiment and one
slot for the session, so it's difficult to make big mistakes in the code.

The worst that can happen is that we forget about registering in some cases
but double registrations are generally okay as long as the proper slot - for
the experiment or the session - is picked up.

Also, the places where we need to modify the context are few, so it seems
this is not going to be a very big burden. It's true that we need to remember
to use the proper dialer, but that's anyway a problem for all tests, to be
sure that we configure dialing correctly.

So, "burn the ships" like Cortez, and entirely remove the pre-existing
byte counting mechanism for Go experiments. We need to keep the callback
for byte counting, though, until we ditch MK experiments.

With this merged, https://github.com/ooni/probe-engine/issues/125 is
mostly done, except that we need to add a bunch of TODOs and follow up
actions after one more round of code inspection.

### DIFF
--- a/experiment.go
+++ b/experiment.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ooni/probe-engine/experiment/web_connectivity"
 	"github.com/ooni/probe-engine/experiment/whatsapp"
 	"github.com/ooni/probe-engine/model"
+	"github.com/ooni/probe-engine/netx/dialer"
 )
 
 const dateFormat = "2006-01-02 15:04:05"
@@ -175,6 +176,7 @@ func newExperimentBuilder(session *Session, name string) (*ExperimentBuilder, er
 
 // Experiment is an experiment instance.
 type Experiment struct {
+	byteCounter   *dialer.ByteCounter
 	callbacks     model.ExperimentCallbacks
 	measurer      model.ExperimentMeasurer
 	report        *collector.Report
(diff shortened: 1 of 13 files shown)


experiment.go
```diff
@@ -270,8 +287,8 @@ type sessionExperimentCallbacks struct {
 }
 
 func (cb *sessionExperimentCallbacks) OnDataUsage(dloadKiB, uploadKiB float64) {
-	cb.sess.kibsReceived.Add(dloadKiB)
-	cb.sess.kibsSent.Add(uploadKiB)
+	cb.sess.byteCounter.Received.Add(dloadKiB)
+	cb.sess.byteCounter.Sent.Add(uploadKiB)
```
bassosimone, line 291: This is a bug. Here we also need to increment the experiment's byte counter!

session.go
```diff
@@ -245,6 +224,7 @@ func (s *Session) NewExperimentBuilder(name string) (*ExperimentBuilder, error)
 // NewOrchestraClient creates a new orchestra client. This client is registered
 // and logged in with the OONI orchestra. An error is returned on failure.
 func (s *Session) NewOrchestraClient(ctx context.Context) (model.ExperimentOrchestraClient, error) {
```
bassosimone, line 226: We need to understand whether operations performed with such client may escape

Issue #506, opened 2020-04-14
ndt7 and dash run over psiphon


bassosimone:
- [x] the ndt7 implementation uses a proxy-aware http.Client/Dialer
- [x] the dash implementation uses a proxy-aware http.Client/Dialer
- [x] we evaluate whether it is correct to forward the IP address anyway (seems it's not)
- [x] a psiphon tunnel is started before starting the experiment
- [x] the data format includes information regarding starting the psiphon tunnel
- [x] possibly create follow-up issues

bassosimone:
All done!

PR #531, opened 2020-04-19
Rework code that manages proxying

bassosimone:
The rationale of this change is the following:

1. honouring `HTTP_PROX`Y is increasingly weird when we have HTTP and non-HTTP pieces of code

2. so, let's instead introduce a concept of explicit proxy at SOCKS5 level

3. if we have a concept of explicit proxy, then use it for everything, to honour the user wishes

4. then we don't need a transport without proxy anymore

Part of https://github.com/ooni/probe-engine/issues/506

### DIFF
--- a/netx/httptransport/httptransport.go
+++ b/netx/httptransport/httptransport.go
@@ -35,16 +35,13 @@ type Resolver interface {
 	LookupHost(ctx context.Context, hostname string) (addrs []string, err error)
 }
 
-// ProxyFunc is the function used to set a proxy.
-type ProxyFunc func(*http.Request) (*url.URL, error)
-
 // Config contains configuration for creating a new transport. When any
 // field of Config is nil/empty, we will use a suitable default.
 type Config struct {
 	ByteCounter *bytecounter.Counter // default: no byte counting
 	Dialer      Dialer               // default: dialer.DNSDialer
 	Logger      Logger               // default: no logging
-	Proxy       ProxyFunc            // default: no proxy
+	ProxyURL    *url.URL             // default: no proxy
 	Resolver    Resolver             // default: system resolver
 	TLSConfig   *tls.Config          // default: attempt using h2
 	TLSDialer   TLSDialer            // default: dialer.TLSDialer
@@ -78,8 +75,12 @@ func New(config Config) RoundTripper {
 		if config.Logger != nil {
 			d = dialer.LoggingDialer{Dialer: d, Logger: config.Logger}
 		}
-		config.Dialer = dialer.DNSDialer{
+		d = dialer.DNSDialer{
 			Resolver: config.Resolver,
+			Dialer:   d,
+		}
+		config.Dialer = dialer.ProxyDialer{
+			ProxyURL: config.ProxyURL,
 			Dialer:   d,
 		}
 	}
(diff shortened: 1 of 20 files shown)


netx/httptransport/httptransport.go
```diff
@@ -78,10 +75,14 @@ func New(config Config) RoundTripper {
...
 		}
-		config.Dialer = dialer.DNSDialer{
+		d = dialer.DNSDialer{
 			Resolver: config.Resolver,
 			Dialer:   d,
 		}
+		config.Dialer = dialer.ProxyDialer{
+			ProxyURL: config.ProxyURL,
+			Dialer:   d,
+		}
```
bassosimone, line 85: We want to put the proxy dialer at the beginning of the chain

session.go
```diff
@@ -44,12 +44,11 @@ type Session struct {
 	availableTestHelpers map[string][]model.Service
 	byteCounter          *bytecounter.Counter
 	httpDefaultTransport httptransport.RoundTripper
-	httpNoProxyTransport httptransport.RoundTripper
 	kvStore              model.KeyValueStore
 	privacySettings      model.PrivacySettings
-	explicitProxy        bool
 	location             *model.LocationInfo
 	logger               model.Logger
+	proxyURL             *url.URL
```
bassosimone, line 51: The transport w/o proxy is gone. The rationale of this change is the following:

1. honouring `HTTP_PROXY` is increasingly weird when we have HTTP and non-HTTP pieces of code

2. so, let's instead introduce a concept of explicit proxy at SOCKS5 level

3. if we have a concept of explicit proxy, then use it for everything, to honour the user wishes

4. then we don't need a transport without proxy anymore

session.go
```diff
@@ -273,6 +256,11 @@ func (s *Session) ProbeIP() string {
 	return ip
 }
 
+// ProxyURL returns the Proxy URL, or nil if not set
+func (s *Session) ProxyURL() *url.URL {
+	return s.proxyURL
+}
+
```
bassosimone, line 263: Rather than telling experiments whether there is an explicit proxy, now we tell them what the proxy is. We cannot use a proxy function, now, because we also have non HTTP consumers, e.g. ndt7.

session.go
```diff
@@ -451,12 +437,14 @@ func (s *Session) maybeLookupLocation(ctx context.Context) (err error) {
 		runtimex.PanicOnError(err, "s.lookupASN #1 failed")
 		cc, err = s.lookupProbeCC(s.CountryDatabasePath(), probeIP)
 		runtimex.PanicOnError(err, "s.lookupProbeCC failed")
-		resolverIP, err = s.lookupResolverIP(ctx)
-		runtimex.PanicOnError(err, "s.lookupResolverIP failed")
-		resolverASN, resolverOrg, err = s.lookupASN(
-			s.ASNDatabasePath(), resolverIP,
-		)
-		runtimex.PanicOnError(err, "s.lookupASN #2 failed")
+		if s.proxyURL == nil {
```
bassosimone, line 440: If we're using a user enforced proxy, then it's not very useful to discover info about a resolver that we don't use and that may actually be misleading if the person looking into the data is not paying attention.

Issue #335, opened 2020-02-14
Use certifi root certificates by default


bassosimone:
Here's what we need to do:

- [ ] here https://github.com/ooni/probe-engine/blob/443e98e8a6c857176ed67cc09e65e8e192c88e30/netx/httptransport/httptransport.go#L138 we need to force the certifi root certificate

- [ ] we need to modify the code in the resources to stop downloading/using cURL's CA

PR #591, opened 2020-05-13
netx: use gocertifi in new code

bassosimone:
Closes https://github.com/ooni/probe-engine/issues/335


bassosimone:
I am going to merge. It seems there is a problem with the new locate API at m-lab that is returning 500 for repeated runs of ndt7 in a non-predictable fashion. I've contacted @stephen-soltesz so that we can look into this together. In any case, this PR can be merged because we have enough checks that are passing the issue does not seem to be related to this PR.

PR #592, opened 2020-05-13
locatev2: experiment without the shaping tag

bassosimone:
The purpose of this experiment is to run the checks using locatev2 without using the shaping code, to single out the possibility that shaping causes the `500` when accessing locatev2.


bassosimone:
So, the failure persist even with the new shaping code disabled. This suggests that the issue is not dependent on whether we have shaping enabled or not. Closing this PR now.

PR #593, opened 2020-05-13
mlablocatev2: understand the failures we have

bassosimone:
The purpose of this experiment is to understand the failures we have when calling mlablocatev2 and the first step is to know all the places where we call the mlablocatev2 API.


bassosimone:
It turns out it's a server side problem, so no further action required here. Closing.

PR #602, opened 2020-05-19
jsonapi, probeservices: refactoring before detecting blocking

bassosimone:
This PR contains a but of unrelated diffs that perform refactoring of jsonapi and probeservices. I want these changes to be in before continuing with https://github.com/ooni/probe/issues/887.


bassosimone:
Merging! (The failure seems to be a false positive and also tests pass locally)

Issue #626, opened 2020-05-25
Sketch out stun_reachability prototype


bassosimone:
As pointed out by https://github.com/backkem in https://github.com/ooni/probe-engine/issues/548#issuecomment-629771669, we can use https://github.com/pion/stun to check for stun reachability of specific servers. The readme of pion/stun includes this code:

```Go
package main

import (
	"fmt"

	"github.com/pion/stun"
)

func main() {
	// Creating a "connection" to STUN server.
	c, err := stun.Dial("udp", "stun.l.google.com:19302")
	if err != nil {
		panic(err)
	}
	// Building binding request with random transaction id.
	message := stun.MustBuild(stun.TransactionID, stun.BindingRequest)
	// Sending request to STUN server, waiting for response message.
	if err := c.Do(message, func(res stun.Event) {
		if res.Error != nil {
			panic(res.Error)
		}
		// Decoding XOR-MAPPED-ADDRESS attribute from message.
		var xorAddr stun.XORMappedAddress
		if err := xorAddr.GetFrom(res.Message); err != nil {
			panic(err)
		}
		fmt.Println("your IP is", xorAddr.IP)
	}); err != nil {
		panic(err)
	}
}
```

which returns your IP address, if the server is reachable. The purpose of this issue is to extend the above code and integrate it into a OONI experiment where input is optional. By default we will use a fixed list of servers (perhaps only `stun.l.google.com:19302` in the initial PoC).

This is part of https://github.com/ooni/probe-engine/issues/243.

bassosimone:
Writing tests required extra time that I borrowed from https://github.com/ooni/probe-engine/issues/614.

PR #630, opened 2020-05-26
Write stun_reachability experiment prototype

bassosimone:
Closes https://github.com/ooni/probe-engine/issues/626

To finish off this PR, needed to make error wrapping code public, which I planned on doing anyway.


bassosimone:
Got a green light to go forward with this in today's internal meeting. Will make sure we have full code coverage and follow-up issues before we're merging.

Issue #621, opened 2020-05-22
{bouncer,collector => probe_services}_base_url


bassosimone:
If we refactor the code to only use probe services, we can implement failover between different data centres by connecting to each data centre and choose the one with the best performance. See https://github.com/ooni/probe-engine/issues/407.

Without this simplifying change in tree, the logic around choosing the bouncer and the collector independently is going to become overcomplicated if coupled with failover between data centres. Because these options are seldom used, and because they are not exposed through the UI, this change is totally fine.

bassosimone:
(Swapped efforts with https://github.com/ooni/probe-engine/issues/407)

bassosimone:
This is now done!

PR #636, opened 2020-05-27
oonimkall: remove support for {Bouncer,Collector}BaseURL

bassosimone:
After this change, we're unable to test that the code WAIs when we
cannot discover backends and open a report.

This will be fixed once we have support for ProbeServicesBaseURL
inside of the session.go file.

Part of https://github.com/ooni/probe-engine/issues/621.


bassosimone:
A reduction in coverage was expected because we can test slightly less than before, as mentioned above. I am going to merge anyway and continue working on this feature.

Issue #407, opened 2020-03-16
Add failover to probes


hellais:
We could potentially open many connections to collectors and just keep the faster one.

Possibly related to: https://github.com/ooni/backend/issues/366.

bassosimone:
This is rather difficult to do until we have the guarantee that we always use the probe services and until we refactor the code to reflect that, removing passing from the bouncer for collectors.

I did https://github.com/ooni/probe-engine/issues/19 instead, which has the same effort estimate.

bassosimone:
To do this, we need to do https://github.com/ooni/probe-engine/issues/621 first. Once that is done, we can modify `session.go·queryBouncer` to attempt fetching TestHelpers from every possible `https` probe service. The fastest probe service would then be chosen. If no `https` probe service is working, we could then fallback to `cloudfronted` bouncers. The choice of the probe service will be saved, and reused for later when open the report, etc.

bassosimone:
Discussed today with @FedericoCeratto and @hellais: we will create N entries like `https://ps$N.ooni.io` (or something similar) and the code will try these entries first. It will choose the one that is fastest to fetch the test helpers. If none of them work, we do fallback like we do now. This strikes a balance between reliability and selecting the most performance probe services endpoint.

bassosimone:
(Swapped efforts with https://github.com/ooni/probe-engine/issues/621)

bassosimone:
Has been implemented. We're still missing DNS names for every data centre. Will discuss this tomorrow in our backend meeting so we're sure we will create follow-up issues.

PR #647, opened 2020-05-29
session: select the fastest probe service

bassosimone:
Part of https://github.com/ooni/probe-engine/issues/407

Includes fixes for a flaky integration test that otherwise prevented clean merging


oonimkall/task_test.go
```diff
@@ -431,7 +431,7 @@ func TestIntegrationInterruptNdt7(t *testing.T) {
 		t.Fatal(err)
 	}
 	go func() {
-		<-time.After(7 * time.Second)
+		<-time.After(11 * time.Second)
```
bassosimone, line 434: This seems to fix an integration test that started failing with GitHub Actions (but not locally)

oonimkall/task_test.go
```diff
@@ -462,8 +462,8 @@ func TestIntegrationInterruptNdt7(t *testing.T) {
 		"status.end",
 		"task_terminated",
 	}
-	if !reflect.DeepEqual(keys, expect) {
-		t.Fatal("seen different keys than expected")
+	if diff := cmp.Diff(expect, keys); diff != "" {
+		t.Fatalf(diff)
```
bassosimone, line 466: This has been useful to understand why an integration test was failing

oonimkall/task_test.go
```diff
@@ -411,8 +411,8 @@ func TestIntegrationInterruptExampleWithInput(t *testing.T) {
 		"status.end",
 		"task_terminated",
 	}
-	if !reflect.DeepEqual(keys, expect) {
-		t.Fatalf("seen different keys than expected: %+v", keys)
+	if diff := cmp.Diff(expect, keys); diff != "" {
+		t.Fatalf(diff)
```
bassosimone, line 415: This has been done to sync this test with another test that we changed to investigate failures

probeservices/client_test.go
```diff
@@ -204,3 +206,385 @@ func TestSortEndpoints(t *testing.T) {
...
+	// put onion first so we also verify that we sort the endpoints
+	in := []model.Service{{
+		Type:    "onion",
+		Address: "httpo://jehhrikjjqrlpufu.onion",
+	}, {
+		Type:    "https",
+		Address: "https://ams-ps.ooni.io",
+	}, {
+		Type:    "https",
+		Address: "https://hkg-ps.ooni.io",
```
hellais, line 255: I would suggest replacing these with another string that is clearly not used, as these domain names are very similar to one that exists, but in fact they do not.

bassosimone: https://github.com/ooni/probe-engine/issues/664

Issue #651, opened 2020-05-29
Use best probe service also for orchestration et al.


bassosimone:
Once https://github.com/ooni/probe-engine/issues/650 is done, we will be using ps.ooni.io as the base URL every time we wanna speak with a probe service. Yet some interactions are currently more robust (e.g. bouncer), while others are more fragile because we hardcode ps.ooni.io. What we should actually be doing is to use the probe service selected in `MaybeLookupBackends` also for these interactions for which we currently hardcode ps.ooni.io.

Closes https://github.com/ooni/probe-engine/issues/63

bassosimone:
Status update! As I wrote in https://github.com/ooni/probe-engine/pull/701#issue-433214370:

> This is half of our journey. We have now the orchestra API inside the probeservices API. But there is more to do after this step:
>
> 1. make sure we can use the selected probe service rather that having code that only works with the default probe service (this may already be working now, but we need tests and confidence it's working)
> 
> 2. do more hammering of the code to make sure everything fits into the right place, we're not exposing code unnecessarily, etc.
>
> 3. make sure we are satisfied by the tests that are currently in place otherwise make sure we write enough tests to have confidence this is WAI
>
> 4. specifically, choose what to do with the URLs API, which now does not even use the client for probe services
>
> All of this will be part of a bunch of subsequent diffs.

bassosimone:
With https://github.com/ooni/probe-engine/pull/715, we're now 90% done. In such PR, I have mentioned a bunch of easy next steps that I think we should look into before calling this issue fully completed. Also, this issue has clearly been XL and I needed logically to work on making the code better and more easily testable. For this reason, I'm adjusting the related milestones.

bassosimone:
Effort corrected. Equivalent number of story points shifted to next sprint.

bassosimone:
With the merging of https://github.com/ooni/probe-engine/pull/718, this work is now complete.

PR #652, opened 2020-05-29
Use ps.ooni.io wherever it makes sense to do so

bassosimone:
Prodded by @hellais. Will give us better performance. Yet, this is just
half of the job. I need to open a follow up issue suggesting that we should
always be using the same probe service that we selected. Otherwise, some
parts of our infra will be less resilient than others.

Such follow up issue is https://github.com/ooni/probe-engine/issues/651.


bassosimone:
It seems the macOS workflow is stuck. I run it locally and it worked.

Issue #646, opened 2020-05-28
Netx research and refactoring in Sprint 15


bassosimone:
Following up from https://github.com/ooni/probe-engine/issues/576

- [x] preliminary work for rewriting `experiment/telegram` using `experiment/urlgetter`
- [x] errorx: make sure we correctly handle `"context canceled"` so it gets scrubbed
- [x] reimplement `experiment/telegram` in terms of `experiment/urlgetter`

This refactoring is functional to measure performance metrics. In turn, by doing that we will be able to measure HTTP fetching performance metrics for Psiphon. This is also a precondition to rewrite Web Connectivity in Go.

bassosimone:
We have now refactored Telegram. More work in next sprint: https://github.com/ooni/probe-engine/issues/684.

PR #683, opened 2020-06-09
telegram: rewrite using experiment/urlgetter

bassosimone:
Part of https://github.com/ooni/probe-engine/issues/646


bassosimone:
Passes all checks from https://github.com/ooni/jafar/pull/28. So, we're good to merge.

PR #700, opened 2020-06-11
internal/orchestra: rewrite to have same constructor as probeservices

bassosimone:
This should be one of the last steps. The orchestra client and the probeservices
client now have basically the same fields and the same constructor.

This means that it should be possible, in a follow-up step, to merge them.

In this diff, I am slightly reducing the code coverage. This will be amended in
a follow-up PR, once I have merged orchestra into probeservices.

Part of https://github.com/ooni/probe-engine/issues/651

### DIFF
--- a/experiment/tor/tor_test.go
+++ b/experiment/tor/tor_test.go
@@ -3,6 +3,7 @@ package tor
 import (
 	"context"
 	"errors"
+	"net/http"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


bassosimone:
We knew coverage was going to be lowered, as mentioned above. Yet, we're doing to merge anyway, since this is an intermediate step towards merging probeservices and orchestra.

(TBH, I didn't expect this process of making orchestra using probeservices to run this long. :-)

Issue #55, opened 2019-09-10
whatsapp: rewrite in Go (1/2)


bassosimone:

bassosimone:
We're mostly done, what is missing TODO is:

- [ ] fail if the registration server returns a non `200` status code
- [ ] fail if the web response is not `200` with the correct title
- [ ] make sure we can pass the `AllEndpoints` option from mobile

After all of this has been done, we can finally call this work complete.

bassosimone:
Done enough for this sprint, will continue with https://github.com/ooni/probe-engine/issues/743 in the next sprint

PR #739, opened 2020-06-30
urlgetter: implement FailOnHTTPError

bassosimone:
In some cases, we want to fail if we see an HTTP status code >= 400.

This is certainly the case when we're doing plain text HTTP.

Both the Telegram implementation and WhatsApp can benefit from that
functionality, when connecting to the web interface.

Thus, I've factored the code out of the Telegram implementation
and I have moved it inside the urlgetter package.

Part of https://github.com/ooni/probe-engine/issues/55

### DIFF
--- a/experiment/telegram/telegram.go
+++ b/experiment/telegram/telegram.go
@@ -69,12 +69,6 @@ func (tk *TestKeys) Update(v urlgetter.MultiOutput) {
 		tk.TelegramWebFailure = v.TestKeys.Failure
 		return
 	}
-	if v.TestKeys.HTTPResponseStatus != 200 {
-		failureString := "http_request_failed" // MK uses it
-		tk.TelegramWebFailure = &failureString
-		tk.TelegramWebStatus = "blocked"
-		return
-	}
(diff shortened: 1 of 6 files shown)


experiment/telegram/telegram.go
```diff
@@ -115,10 +118,10 @@ func (m measurer) Run(ctx context.Context, sess model.ExperimentSession,
 		{Target: "http://149.154.167.91:443/", Config: urlgetter.Config{Method: "POST"}},
 		{Target: "http://149.154.171.5:443/", Config: urlgetter.Config{Method: "POST"}},
 
-		{Target: "http://web.telegram.org/", Config: urlgetter.Config{Method: "GET"}},
-		{Target: "https://web.telegram.org/", Config: urlgetter.Config{Method: "GET"}},
+		{Target: "http://web.telegram.org/", Config: urlgetter.Config{FailOnHTTPError: true}},
+		{Target: "https://web.telegram.org/", Config: urlgetter.Config{FailOnHTTPError: true}},
```
bassosimone, line 122: "GET" is the default

Issue #684, opened 2020-06-09
experiment/sni_blocking: rewrite using urlgetter


bassosimone:
Following up from https://github.com/ooni/probe-engine/issues/646

- [ ] reimplement `experiment/sni_blocking` in terms of `experiment/urlgetter`

This refactoring is functional to measure performance metrics. In turn, by doing that we will be able to measure HTTP fetching performance metrics for Psiphon. This is also a precondition to rewrite Web Connectivity in Go.

bassosimone:
This task has been completed!

PR #744, opened 2020-07-01
sniblocking: rewrite using urlgetter

bassosimone:
A more substantial rewrite would have been possible. But let's proceed
piecemeal and, for now, just stop depending on the old netx.

This was the objective and we achieved it.

When we'll tackle the task of improving testing, we'll also refactor
the tests and see if we can also simplify sniblocking further.

Part of https://github.com/ooni/probe-engine/issues/684

### DIFF
--- a/experiment/sniblocking/sniblocking.go
+++ b/experiment/sniblocking/sniblocking.go
@@ -115,9 +115,11 @@ func (m *measurer) measureone(
 	case <-time.After(sleeptime):
 	case <-ctx.Done():
 		s := modelx.FailureInterrupted
+		failedop := modelx.TopLevelOperation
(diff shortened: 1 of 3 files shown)


experiment/sniblocking/sniblocking.go
```diff
@@ -132,39 +116,33 @@ func (m *measurer) measureone(
 	case <-ctx.Done():
 		s := modelx.FailureGenericTimeoutError
 		return Subresult{
-			Failure: &s,
-			SNI:     sni,
+			TestKeys: urlgetter.TestKeys{
+				Failure: &s,
+			},
```
bassosimone, line 121: Missing THAddress here

PR #760, opened 2020-07-10
use Powerdns as default DoH

### LINKED ISSUE #759: sessionresolver: use powerdns

xhdix:
Because Powerdns is better in terms of GDPR!

(I tried to make changes in some places in a fair way. I don't know if I did too much or too little!)

Closes #759

### DIFF
--- a/experiment/urlgetter/configurer_test.go
+++ b/experiment/urlgetter/configurer_test.go
@@ -70,6 +70,82 @@ func TestConfigurerNewConfigurationVanilla(t *testing.T) {
 	}
 }
 
+func TestConfigurerNewConfigurationResolverDNSOverHTTPSPowerdns(t *testing.T) {
+	saver := new(trace.Saver)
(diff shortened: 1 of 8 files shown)


internal/oonitemplates/oonitemplates.go
```diff
@@ -173,6 +173,10 @@ func configureDNS(seed int64, network, address string) (modelx.DNSResolver, erro
 		return nil, err
 	}
 	fallbacks := []dnsFallback{
+		dnsFallback{
+			network: "doh",
+			address: "https://doh.powerdns.org/",
+		},
```
bassosimone, line 179: Can you remove this part of the diff? The oonitemplates.go code is code that I am removing and it does not matter adding this bit of code here. Now only one test out of many is still using this code.

xhdix: done! :D

xhdix:
I'm glad I could help ;)

PR #772, opened 2020-07-11
DoH: better timeouts and TRR2-like implementation

xhdix:
Let us use 15 seconds as timeout for connecting. Let us give DoH 45 seconds to resolve by default. This means, if there are timeouts, we can attempt with three IP addresses. At the same time, implement something like TRR2 for the session resolver such that we fallback to using the system resolver quite quickly if DoH doesn't work.


bassosimone:
Thanks @xhdix, I'm going to merge now!

Issue #776, opened 2020-07-13
Rewrite Web Connectivity in Go (1/n): write the code


bassosimone:
Part of https://github.com/ooni/probe-engine/issues/40

bassosimone:
This is done! Next step is QA: https://github.com/ooni/probe-engine/issues/810

PR #778, opened 2020-07-13
Rewrite Web Connectivity in Go (1/n)

bassosimone:
This diff implements the basic machinery. We perform the measurement
and the control measurement. We'll write tests next.

For now, we still use MK for Web Connectivity. We will not switch to the
new implementation until it's feature complete and tested.

Part of https://github.com/ooni/probe-engine/issues/776.


bassosimone:
Coverage decreased because this code is not tested. I'll add tests in the next PR.

Issue #829, opened 2020-07-23
jafar: write QA for IM and middlebox tests


bassosimone:

bassosimone:
done!

PR #830, opened 2020-07-24
QA: implement whatsapp checks

bassosimone:
Part of https://github.com/ooni/probe-engine/issues/829

### DIFF
--- a/QA/telegram.py
+++ b/QA/telegram.py
@@ -1,17 +1,11 @@
 #!/usr/bin/env python3
 
 
-""" ./qa/telegram.py - main QA script for telegram
+""" ./QA/telegram.py - main QA script for telegram
 
     This script performs a bunch of telegram tests under censored
     network conditions and verifies that the measurement is consistent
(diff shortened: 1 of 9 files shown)


QA/telegram.py
```diff
@@ -1,18 +1,12 @@
...
     This script performs a bunch of telegram tests under censored
     network conditions and verifies that the measurement is consistent
     with the expectations, by parsing the resulting JSONL. """
 
-# TODO(bassosimone): I sometimes see tests failing with eof_error, which is
-# probably caused by me using a mobile connection. I believe we should attempt
-# to be strict here, because we're doing QA. But maybe I'm wrong?
-#
-# For now, I'm adding a bunch of sleeps to mitigate.
-
```
bassosimone, line 15: This comment was outdated: I have removed fine grained checks to mitigate this issue. Otherwise, it was very difficult to get consistent results from repeated run. Too many false positives are not good here.

Issue #579, opened 2020-05-04
Don't label site "accessible" if there is a control failure, specially not if really inaccesoble


andresazp:
## Expected Behavior

If a measurement has a control failure, it shouldn't be label as accessible; especially if the probe couldn't load it

## Actual Behavior

I'm still checking, but in some cases were the experiment DNs results in no answers (typical VE censorship) and there is a DNS_lookup_error in the control.

In this example a site is possibly DNS-blocked, but DNS control failure leads the current logic: https://explorer.ooni.org/measurement/20200504T045941Z_AS8048_jtDBVHr1LbzO92dFWQ2rgUTadiWfSgLmNIQ55T4a8CpciJWiek?input=https://saludvzla.com/

Perhaps a new category, similar to how it fails in other conditions, should be introduced; Or like on the explorer it says "Ok", instead of "Accesible"

**Update:** after reviewing form different points, indeed the domain is not giving any answers, so there wan't a problem with the control per se,  this example doesn't reflect an error. 

But it might still be reasonable to label such as response as "non inconsistent". If there are problems with the dns control or answers from control are manipulated, then it shouldn't necessarily be assumed that the site is accesible, even if the fail is the same. 

Furthermore if there is an active HTTP block in place that couldn't be verified beacuse of a temporary DNS outage, perhaps the best is not to have it labeled as "accesible", again, no contact to the site was performed.

## Specifications

  - Version:
ooni probe 2.3.0
measurement kit 0.10.11
  - Platform: iOs, presumably all platforms as it's part of the measurement (?)

bassosimone:
> @bassosimone what do you think of this?

Web Connectivity is working as intended in this example. As @hellais says, the domain does not exist, therefore the control and the probe are currently consistent.

At the same time, the Web Connectivity model needs to be improved, in my view. I can say that I sympathise with @andresazp concerns here. It seems to me that a good high level objective to pursue is to make the measurement result more obvious to understand.

On a more strategic standpoint, I am not able to say how. I have not spent enough time thinking on a comprehensive strategy to make this happen. (I am thinking about this, but I'd rather not write down my incomplete thoughts, and I'd rather spend more time thinking on that.)

[...]

PR #842, opened 2020-08-13
QA/webconnectivity.py: add test case for nonexistent domain

bassosimone:
The behavior of probe-engine is better than the one of MK. We have
disabled this check for MK and we should open an issue.

The following diff shows the difference in the toplevel keys:

```diff
--- MK.jsonl	2020-08-13 13:12:05.741791884 +0200
+++ OPE.jsonl	2020-08-13 13:12:14.092840549 +0200
@@ -1,9 +1,9 @@
 {
   "input": "http://antani.xyz",
   "test_keys": {
-    "accessible": true,
+    "accessible": null,
     "agent": "redirect",
-    "blocking": false,
+    "blocking": null,
     "body_length_match": null,
     "body_proportion": 0,
     "client_resolver": "91.80.36.88",
@@ -16,15 +16,16 @@
         "body_length": -1,
         "failure": "dns_lookup_error",
         "headers": {},
-        "status_code": -1
+        "status_code": -1,
+        "title": ""
       },
       "tcp_connect": {}
     },
     "control_failure": null,
     "dns_consistency": "consistent",
-    "dns_experiment_failure": null,
+    "dns_experiment_failure": "dns_nxdomain_error",
     "headers_match": null,
-    "http_experiment_failure": "dns_lookup_error",
+    "http_experiment_failure": null,
     "retries": null,
     "socksproxy": null,
     "status_code_match": null,
```

Because `antani.xyz` does not exist, OPE is clearly right here. I am also
wondering whether this issue with MK is related to the following issue:

https://github.com/ooni/probe-engine/issues/579

Maybe?

What I find rather interesting, also, is that there is no indication of
"dns_experiment_failure" in MK even though it clearly failed. Yet, we see
consistency in the "dns_consistency" key. But, still, I would argue it's
not so smart to elide the DNS failure in this case.

Also, note that the HTTP experiment is not even performed by OPE because
we really don't know what to do without IP addresses for the domain.

### DIFF
--- a/QA/webconnectivity.py
+++ b/QA/webconnectivity.py
@@ -29,39 +29,41 @@ def execute_jafar_and_return_validated_test_keys(
     tk = common.execute_jafar_and_miniooni(
         ooni_exe, outfile, experiment_args, tag, args
     )
-    assert isinstance(tk["requests"], list)
-    assert len(tk["requests"]) > 0
-    for entry in tk["requests"]:
-        assert isinstance(entry, dict)
-        failure = entry["failure"]
-        assert isinstance(failure, str) or failure is None
-        assert isinstance(entry["request"], dict)
-        req = entry["request"]
-        common.check_maybe_binary_value(req["body"])
-        assert isinstance(req["headers"], dict)
-        for key, value in req["headers"].items():
-            assert isinstance(key, str)
-            common.check_maybe_binary_value(value)
-        assert isinstance(req["method"], str)
-        assert isinstance(entry["response"], dict)
-        resp = entry["response"]
-        common.check_maybe_binary_value(resp["body"])
-        assert isinstance(resp["code"], int)
-        if resp["headers"] is not None:
-            for key, value in resp["headers"].items():
+    if tk["requests"]:
+        assert isinstance(tk["requests"], list)
+        assert len(tk["requests"]) > 0
+        for entry in tk["requests"]:
+            assert isinstance(entry, dict)
+            failure = entry["failure"]
+            assert isinstance(failure, str) or failure is None
+            assert isinstance(entry["request"], dict)
+            req = entry["request"]
+            common.check_maybe_binary_value(req["body"])
+            assert isinstance(req["headers"], dict)
+            for key, value in req["headers"].items():
                 assert isinstance(key, str)
                 common.check_maybe_binary_value(value)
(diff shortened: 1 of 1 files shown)


QA/webconnectivity.py
```diff
@@ -29,39 +29,41 @@ def execute_jafar_and_return_validated_test_keys(
...
-            assert isinstance(key, str)
-            common.check_maybe_binary_value(value)
-        assert isinstance(req["method"], str)
-        assert isinstance(entry["response"], dict)
-        resp = entry["response"]
-        common.check_maybe_binary_value(resp["body"])
-        assert isinstance(resp["code"], int)
-        if resp["headers"] is not None:
-            for key, value in resp["headers"].items():
+    if tk["requests"]:
```
bassosimone, line 32: `tk["requests"]` is `nil` here with OPE

QA/webconnectivity.py
```diff
@@ -29,39 +29,41 @@ def execute_jafar_and_return_validated_test_keys(
...
+            assert isinstance(req["method"], str)
+            assert isinstance(entry["response"], dict)
+            resp = entry["response"]
+            common.check_maybe_binary_value(resp["body"])
+            assert isinstance(resp["code"], int)
+            if resp["headers"] is not None:
+                for key, value in resp["headers"].items():
+                    assert isinstance(key, str)
+                    common.check_maybe_binary_value(value)
+    if tk["tcp_connect"]:
```
bassosimone, line 55: `tk["tcp_connect"]` is `nil` here with OPE

bassosimone:
So, after reading https://github.com/ooni/probe-engine/issues/579 more carefully, the test case we've introduced here is indeed the one originally reported by @andresazp. With OPE, this measurement is flagged as failed. Which is better than saying that all is good, because definitely it's not true that all is good here. I also think that the original suggestion by @andresazp is correct: we should treat this case specially. My plan is actually to first finish making sure we don't do worst than MK and then start addressing these corner cases by defining specific keys for identifying them.

bassosimone:
So, to recap, I'm merging this PR where the case in which the website is down is flagged as failed measurement. I think this is better than flagging as "all good" because users will be confused by that. The proper solution is to extend our mechanism of classification to also encompass this specific case, but that is something to do after the QA process is over.

PR #873, opened 2020-08-24
fix(webconnectivity/summary): logging: add missing colon

bassosimone:
Discovered when smoke testing v3.0.7-beta.


bassosimone:
Merging notwithstanding the failing check. It seems there is a (temporary?) issue with the stunreachability experiment in GitHub Actions that I cannot reproduce locally. I am going to consider this a temporary glitch for now.

Issue #893, opened 2020-08-28
oonimkall: design and implement geolocate and resubmit APIs


bassosimone:
The DESIGN.md document currently does not include plans for replacing geolocate and resubmit APIs. The implementation of such APIs in measurement-kit/mkall-ios and measurement-kit/android-libs is rather limited. Specifically, it is a bit backwards that logging stores the logs into a string and that it is not possible to easily interrupt geolocation. These limitations are grounded into the difficulty of writing C++ code compared to writing Go code and using `go mobile`. Therefore we can do better. [Edit: in all fairness we could have used `swig` to possibly ease our burden a bit.]

In Sprint 20, I started designing Go based new APIs for that. The core design choice here was to expose Go objects directly to Java/ObjC code, rather than implementing a set of JSON based wrappers like before. The core reason is _because we can_ and we are rather confident that this would lead us to better the interaction between probe-engine and the mobile apps.

*Edit*:

* final design: https://github.com/ooni/probe-engine/issues/893#issuecomment-689031449

* follow-up actions: https://github.com/ooni/probe-engine/issues/893#issuecomment-685688006

bassosimone:
## A new file: oonimkall/session.go

The new API proposal adds a new file to `oonimkall`. The file is named `session.go` because we need to expose a session to the programmer to (mainly) take advantage of network connection caching.

```Go
// file: oonimkall/session.go

package oonimkall

import (
	"context"
	"encoding/json"
	"errors"
	"fmt"
	"reflect"
	"sync"

	engine "github.com/ooni/probe-engine"
	"github.com/ooni/probe-engine/model"
	"github.com/ooni/probe-engine/probeservices"
)
```

As you can see above, the set of imports is rather limited, but could grow in the future if we also decide to support running tests through this interface (out of scope for now).

bassosimone:
## Allowing Java/ObjC to provide a logger

The following code allows mobile apps to provide us with an implementation of a Logger. Because we don't want to cause too many cross language calls, we also added code to ensure `Debug` and `Debugf` are cross calls _only_ when we really need to dispatch those logs somewhere interesting.

```Go
// SessionLogger is the logger used by a session.
type SessionLogger interface {
	Debug(msg string)
	Info(msg string)
	Warn(msg string)
}

type sessionLoggerVerbose struct {
	SessionLogger
}

func (slv sessionLoggerVerbose) Debugf(format string, v ...interface{}) {
[...]

PR #913, opened 2020-09-04
probeservices: add test_start_time to open-report request

bassosimone:
The field is deprecated but not using it is going to change the
resubmitter behavior wrt MK's submitter.

So, let's be pragmatic and add this field again.

See https://github.com/ooni/probe-engine/issues/893#issuecomment-686973455


bassosimone:
Merging with recurrent qa/fbmessenger failure (see https://github.com/ooni/probe-engine/issues/906).

Issue #770, opened 2020-07-11
webconnectivity: understand DNS false positives (1/n)


bassosimone:
@FedericoCeratto provided me with a list of tests that seem false positives. While doing QA, it makes sense to take a look at them. It may be that they could help us to fix the code and/or write better tests.

bassosimone:
Moving to next sprint; during this sprint, I worked on https://github.com/ooni/probe-engine/issues/894 instead.

bassosimone:
Started working on this. I've looked into roughly 1/5 of the measurements. I used code at https://github.com/ooni/probe-engine/pull/931 as a prototype to start understanding what is happening in such measurements. I'm done for this sprint but before closing this issue I will need to create a new follow-up issue for continuing this work.

PR #931, opened 2020-09-10
cmd/webscore: rescore previous web measurements with the Go engine

bassosimone:
This diff introduces a new PoC where we score previous measurements using
the scoring mechanism in the OONI engine. This should in principle allow us
to evaluate whether we're able to detect and workaround old false positive
conditions. The main trouble is that past measurements in many cases have
less fields than now. So, what I have done in this diff is to just recompute
the overall result w/o recomputing keys like `dns_consistent`.

This exercise has been useful to bootstrap an investigation on whether we
could make sense of a bunch of anomalies, as documented by

https://github.com/ooni/probe-engine/issues/770

More works needs to be done to continue this research in future Sprints.


bassosimone:
We're not accepting pull requests in this repository.

PR #936, opened 2020-09-11
Ensuring we can handle IDNs without issues

xhdix:
Further study:
https://developer.mozilla.org/en-US/docs/Mozilla/Internationalized_domain_names_support_in_Mozilla
https://docs.microsoft.com/en-us/windows/win32/intl/handling-internationalized-domain-names--idns


bassosimone:
Thank you, meged!

PR #937, opened 2020-09-11
add pull_request to golang.yml

xhdix:


bassosimone:
It seems #936 also caused issues with the build. I suggest we work together to fix the build as part of this PR!

Issue #926, opened 2020-09-09
miniooni: allow reading input for an experiment from a file


bassosimone:
* a good mnemonic for that is probably `-f` and/or `--file`

* it's probably fine, for the first implementation, to read all the input at once

Multiple people have expressed interest into this feature either in `miniooni` or in `probe-cli`. Considering we'll expose `miniooni` as a subcommand for `probe-cli`, this could actually be quite useful for many.

bassosimone:
This has clearly been done!

PR #939, opened 2020-09-15
libminiooni: allow newline at EOF with -f/--file

bassosimone:
It didn't occur to me initially, but the fact that vim adds a newline at
the end of the file adds an extra blank input to be tested.

We actually don't want to emit an error for such an input and the fix
is just to avoid adding blank inputs to the inputs list.

Since the fix is so easy, while there allow any other blank line in
the middle of the file, since it costs us nothing.

Part of https://github.com/ooni/probe-engine/issues/926

### DIFF
--- a/libminiooni/libminiooni.go
+++ b/libminiooni/libminiooni.go
@@ -237,8 +237,14 @@ func loadFileInputs(opts *Options) {
 		}
 		content, err := ioutil.ReadFile(opts.InputFilePath)
 		fatalOnError(err, "cannot read input file")
-
-		opts.Inputs = strings.Split(string(content), "\n")
+		// Implementation note: when you save file with vim, you have newline at
+		// end of file and you don't want to consider that an input line. While there
+		// ignore any other empty line that may occur inside the file.
+		for _, input := range strings.Split(string(content), "\n") {
+			if input != "" {
+				opts.Inputs = append(opts.Inputs, input)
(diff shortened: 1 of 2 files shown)


bassosimone:
Tests are failing because the testing probe services are down. 😿

bassosimone:
So, the tests failure is clearly caused by

```
% ping 37.218.241.86
PING 37.218.241.86 (37.218.241.86): 56 data bytes
Request timeout for icmp_seq 0
92 bytes from ip4-104-245-124-114.rdns.netrouting.net (104.245.124.114): Destination Host Unreachable
Vr HL TOS  Len   ID Flg  off TTL Pro  cks      Src      Dst
 4  5  48 5400 3fbb   0 0000  31  01 70c8 192.168.1.5  37.218.241.86 

92 bytes from ip4-104-245-124-114.rdns.netrouting.net (104.245.124.114): Destination Host Unreachable
Vr HL TOS  Len   ID Flg  off TTL Pro  cks      Src      Dst
 4  5  48 5400 673e   0 0000  31  01 4945 192.168.1.5  37.218.241.86 
```

I run the relevant tests locally and they are passing:

```
% go test ./libminiooni/...
ok  	github.com/ooni/probe-engine/libminiooni	12.777s
```

Nothing else has been changed. I think it is fine to merge anyway.

Issue #930, opened 2020-09-10
ndt7: upgrade to locate v2 API


bassosimone:
From @stephen-soltesz: "/v2beta1/query -> /v2/nearest"

bassosimone:
done!

stephen-soltesz:
@bassosimone I know and understand that it won't be a priority to package a new release now. Is there an issue that I can track for it when it does happen? This will allow me to retire the v2beta1 target once it has happened.

bassosimone:
@stephen-soltesz sure thing! So, you should track:

1. probe-ios, probe-android: release v2.7.0: https://github.com/ooni/probe/issues/1251

2. OONI Probe Desktop 3.0.4: https://github.com/ooni/ooni.org/issues/600

My guesstimate is that these releases won't happen before 15th October.

bassosimone:
@stephen-soltesz FYI: we use staged rollouts, so it takes ~7 days after the release date for the mobile clients fleet to be completely updated to a new version. This means it's IMHO safe to pull the plug of v2beta.1 2-3 weeks after release date (computed as when the related issue is closed).

PR #940, opened 2020-09-15
internal/mlablocatev2: use v2/nearest API

bassosimone:
See https://github.com/ooni/probe-engine/issues/930


bassosimone:
Safe to merge because failures are caused by issues in `mia-ps-test.ooni.nu`.

PR #943, opened 2020-09-18
add IDN resolver

### LINKED ISSUE #938: netx/resolver: IDN not supported when using miekg/dns
As we learned in https://github.com/ooni/probe-engine/pull/936 and https://github.com/ooni/probe-engine/commit/5a373e6b5a8067e28096dada0966d09e7718d0c2, `netx/resolver` does not support IDN when we are using the `miekg/dns` based resolvers as opposed to Golang's resolver. We should try and improve upon that.

xhdix:
Fix #938

### DIFF
--- a/netx/netx.go
+++ b/netx/netx.go
@@ -136,7 +136,7 @@ func NewResolver(config Config) Resolver {
 		r = resolver.SaverResolver{Resolver: r, Saver: config.ResolveSaver}
 	}
 	r = resolver.AddressResolver{Resolver: r}
-	return r
+	return resolver.IDNAResolver{Resolver: r}
 }
 
 // NewDialer creates a new Dialer from the specified config
--- a/netx/netx_test.go
+++ b/netx/netx_test.go
@@ -19,7 +19,11 @@ import (
 
 func TestNewResolverVanilla(t *testing.T) {
 	r := netx.NewResolver(netx.Config{})
-	ar, ok := r.(resolver.AddressResolver)
+	ir, ok := r.(resolver.IDNAResolver)
+	if !ok {
+		t.Fatal("not the resolver we expected")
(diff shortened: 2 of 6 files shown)


bassosimone:
@xhdix would updating and merging master into this PR? I've just committed a change so that we run more comprehensive checks opened by contributors, and I'd like to see these checks running before proceeding with review. Thanks!

xhdix:
It's done. Need fixup?

bassosimone:
> It's done.

Thanks!

> Need fixup?

No, it's fine, I am going to squash and merge anyway

bassosimone:
@xhdix I was looking now at the coverage, which is decreasing. It seems we are missing a check where the conversion for IDNA domain names fails. Please, see the lines that are not covered:

https://coveralls.io/jobs/67999399/source_files/2428426956.

Based on the implementation, it seems to me we should be able to provoke an error by controlling the overall length of the domain name (see https://github.com/golang/net/blob/62affa334b73/idna/idna10.0.0.go#L387).

xhdix:
> Based on the implementation, it seems to me we should be able to provoke an error by controlling the overall length of the domain name (see https://github.com/golang/net/blob/62affa334b73/idna/idna10.0.0.go#L387).

So I mistakenly thought this error should suffice:
https://github.com/ooni/probe-engine/blob/9d31430034e6b2bbfc8d0b08dc3b368a65e1f35e/netx/resolver/idna_test.go#L72

:thinking: 
Thanks.

bassosimone:
> So I mistakenly thought this error should suffice:

Let me take a more in depth look at it.

bassosimone:
Closing in favour of https://github.com/ooni/probe-engine/pull/952

Issue #951, opened 2020-09-24
oonimkall/session: redesign for simplicity


bassosimone:
[Edit: turns out renaming was not so useful and a redesign for simplicity was better; see  below for more details.]

We discussed this with @lorenzoPrimi when working on https://github.com/ooni/probe-android/pull/360. There is a bunch of renames that could significantly improve the readability of such an API.

- [ ] Session::newMakeSubmitterTask => Session::newProbeServicesClient
- [ ] MakeSubmitterTask => ProbeServicesClient
- [ ] MakeSubmitterTask::run -> ProbeServicesClient::openReport
- [ ] Submitter => Report

With these changes, it seems we can call the task of writing an API for mobile done and we can move forward with making a new stable release of ooni/probe-engine (see https://github.com/ooni/probe-engine/issues/871).

bassosimone:
## Redesigning oonimkall/session.go for simplicity

As I wrote in https://github.com/ooni/probe-engine/commit/0ac95dc3871e5f8a12933715b0cb7079d89f2f67:

> When helping @lorenzoPrimi with ooni/probe-android#360
> I had the impression there was too much complexity.
> 
> Then we said we could improve the situation by changing the class names
> as written in #951 (comment).
> 
> But then there was still too much complexity. I could not easily picture in
> my mind why each class was needed. This was actually a strong signal that we
> needed to refactor the code more than just renaming clases.
> 
> So, here's the new approach (for which I'll write soon a mini design document
> that will be referenced inside the code - or maybe this is the document!):
> 
> 1. Session uses runtime.SetFinalizer therefore we don't need to be concerned
> about wrapping it and using AutoCloseable in Java. It is not normal for Go code
> to use finalizers, but I believe it makes sense in this context, _because_ it
> simplifies interfacing Go and Java code.
> 
> 2. We introduce again a Context instance _because_ it was too painful to have
> each Task define the same operations and also too bloated.
> 
> 3. A Context is created from a Session, therefore the Session can keep track
> of what `cancel()` functions to be called when done. So now the programmer is
> feee to just not be concerned about calling `cancel()` _unless_ there is the
> need for really cancelling the operation. The Session will specifically call
> all the cancelling functions inside of its finalizer.
> 
> 4. All Session methods are now "synchronized" with respect to a mutex that
> is inside of the session itself. This removes any concerns regarding the usage
> from multiple threads. Yet, we're still not able to have a long running
> session yet, mainly because now we're tracking child contexts. But I also
[...]

PR #955, opened 2020-09-29
oonimkall: expose method to update resources

bassosimone:
See https://github.com/ooni/probe-engine/issues/951#issuecomment-700625131


bassosimone:
Merging with known qa/fbmessenger glitch (https://github.com/ooni/probe-engine/issues/906).

PR #958, opened 2020-09-30
Http host based blocking experiment

### LINKED ISSUE #953: HTTP host-based blocking experiment
A member of the community is working to implement the HTTP host-based experiment in Go. This issue tracks the time I will spent providing advices, performing code reviews, and helping to land this effort.

62w71st:
http host experiment. 

Towards #953

### DIFF
--- a/allexperiments.go
+++ b/allexperiments.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ooni/probe-engine/experiment/fbmessenger"
 	"github.com/ooni/probe-engine/experiment/hhfm"
 	"github.com/ooni/probe-engine/experiment/hirl"
+	"github.com/ooni/probe-engine/experiment/httphostheader"
 	"github.com/ooni/probe-engine/experiment/ndt7"
 	"github.com/ooni/probe-engine/experiment/psiphon"
 	"github.com/ooni/probe-engine/experiment/sniblocking"
@@ -136,6 +137,18 @@ var experimentsByName = map[string]func(*Session) *ExperimentBuilder{
 			},
 			config:      &hhfm.Config{},
 			inputPolicy: InputNone,
+		}
+	},
+
+	"http_host_header": func(session *Session) *ExperimentBuilder {
+		return &ExperimentBuilder{
+			build: func(config interface{}) *Experiment {
+				return NewExperiment(session, httphostheader.NewExperimentMeasurer(
+					*config.(*httphostheader.Config),
+				))
+			},
+			config:      &httphostheader.Config{},
+			inputPolicy: InputRequired,
 		}
 	},
 
--- a/experiment/httphostheader/httphostheader.go
+++ b/experiment/httphostheader/httphostheader.go
@@ -1,0 +1,80 @@
+// Package httphost contains the httphost network experiment.
+//
+// This experiment has not been specified yet. It is nonetheless available for testing
+// and as a building block that other experiments could reuse.
+
+package httphostheader
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/ooni/probe-engine/experiment/urlgetter"
+	"github.com/ooni/probe-engine/model"
+)
+
+const (
+	testName    = "http_host_header"
+	testVersion = "0.1.0"
+)
+
+// Config contains the experiment config.
+type Config struct {
+	// TestHelperURL is the address of the test helper.
+	TestHelperURL string
+}
+
+// TestKeys contains httphost test keys.
+type TestKeys struct {
+	urlgetter.TestKeys
+	THAddress string `json:"th_address"`
+}
+
+// Measurer performs the measurement.
+type Measurer struct {
+	config Config
+}
+
+// ExperimentName implements ExperimentMeasurer.ExperiExperimentName.
+func (m *Measurer) ExperimentName() string {
+	return testName
+}
+
+// ExperimentVersion implements ExperimentMeasurer.ExperimentVersion.
+func (m *Measurer) ExperimentVersion() string {
+	return testVersion
+}
+
+// Run implements ExperimentMeasurer.Run.
+func (m *Measurer) Run(
+	ctx context.Context,
+	sess model.ExperimentSession,
+	measurement *model.Measurement,
+	callbacks model.ExperimentCallbacks,
+) error {
+	if measurement.Input == "" {
+		return errors.New("Experiment requires measurement.Input")
+	}
+	if m.config.TestHelperURL == "" {
+		m.config.TestHelperURL = "http://www.example.com"
+	}
+	urlgetter.RegisterExtensions(measurement)
+	// perform the measurement
+	g := urlgetter.Getter{
+		Begin: measurement.MeasurementStartTimeSaved,
+		Config: urlgetter.Config{
+			HTTPHost: string(measurement.Input),
+		},
+		Session: sess,
+		Target:  fmt.Sprintf(m.config.TestHelperURL),
+	}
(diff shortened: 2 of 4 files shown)


62w71st:
@bassosimone Applied comments. PTAL.

62w71st:
Working on resolving the conflicts in `experiment.go`

62w71st:
All tests pass locally:
```
ok  	github.com/ooni/probe-engine	137.455s
```

62w71st:
Mm found more failing tests from the PR test pipeline. Working on fixing them ...

bassosimone:
@62w71st ACK, I will take a look at this early this week! Thanks!

experiment/httphost/httphost.go
```diff
@@ -0,0 +1,94 @@
+// Package httphost contains the httphost network experiment.
+//
```
bassosimone, line 2:
```suggestion
//
// This experiment has not been specified yet. It is nonetheless available for testing
// and as a building block that other experiments could reuse.
```

experiment/httphost/httphost.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/ooni/probe-engine/experiment/urlgetter"
+	"github.com/ooni/probe-engine/model"
+)
+
+const (
+	testName    = "http_host_header"
```
bassosimone, line 16: Would you mind renaming the directory and the files `httphostheader`? I feel like this is a nice way to work around the fact that there is a legacy, much-more-complex experiment called `http_host`. Thank!

experiment/httphost/httphost.go
```diff
@@ -0,0 +1,94 @@
...
+	if measurement.Input == "" {
+		return errors.New("Experiment requires measurement.Input")
+	}
+	if m.config.TestHelperURL == "" {
+		return errors.New("Experiment requires test helper address.")
+	}
+	urlgetter.RegisterExtensions(measurement)
+	// perform the measurement
+	g := urlgetter.Getter{
+		Begin: measurement.MeasurementStartTimeSaved,
```
bassosimone, line 71: I like this idea of getting `Begin` from the measurement directly, thanks!

allexperiments.go
```diff
@@ -139,6 +140,20 @@ var experimentsByName = map[string]func(*Session) *ExperimentBuilder{
...
+	"http_host_header": func(session *Session) *ExperimentBuilder {
+		return &ExperimentBuilder{
+			build: func(config interface{}) *Experiment {
+				return NewExperiment(session, httphost.NewExperimentMeasurer(
+					*config.(*httphost.Config),
+				))
+			},
+			config: &httphost.Config{
+				TestHelperURL: "http://example.com",
+			},
```
bassosimone, line 152:
```suggestion
			config: &httphost.Config{},
```

As stated below, I'd rather configure the default inside the experiment directory itself.

bassosimone (changes requested):
Thank you so much @62w71st for stepping up and helping us improve writing preliminary checks that could be useful to determine we probably don't need to connect to a website, because we already determined that there are selected forms of censorship. This work is really integral to a more comprehensive strategy to improve the Web Connectivity experiment.

In this review I limited myself to looking into the code and have provided you with suggestions pertaining to the style and the code that should make it okay for merging. I will invest some more time during the next two weeks to look into the tests, run the code locally, and let you know whether there are extra changes to apply in your testing code.

I have a feeling this work is landing and we'll have merged this PR by 1-1.5 weeks.

Thanks again! :guitar:

experiment/httphost/httphost.go
```diff
@@ -0,0 +1,95 @@
...
+type Subresult struct {
+	urlgetter.TestKeys
+	THAddress string `json:"th_address"`
+}
+
+// TestKeys contains httphost test keys.
+type TestKeys struct {
+	Control Subresult `json:"control"`
+	Result  string    `json:"result"`
+	Target  Subresult `json:"target"`
```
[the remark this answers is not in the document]
bassosimone: Ah, I just realised maybe `th_address` is better named `th_url`!

experiment/httphost/httphost.go
```diff
@@ -0,0 +1,95 @@
...
+	ctx context.Context,
+	sess model.ExperimentSession,
+	measurement *model.Measurement,
+	callbacks model.ExperimentCallbacks,
+) error {
+	if measurement.Input == "" {
+		return errors.New("Experiment requires measurement.Input")
+	}
+	if m.config.TestHelperAddress == "" {
+		return errors.New("Experiment requires test helper address.")
```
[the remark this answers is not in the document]
62w71st: good point on making httptest module self-contained.

experiment/httphost/httphost.go
62w71st: I think I might have just copied it from elsewhere  ... but thanks :)

62w71st (commented):
Thanks for the comments! Applied suggested changes. PTAL

62w71st:
Please hold on the review -- still working on this. Missed some of the comments.

experiment/httphost/httphost.go
```diff
@@ -0,0 +1,95 @@
...
+	g := urlgetter.Getter{
+		Begin: measurement.MeasurementStartTimeSaved,
+		Config: urlgetter.Config{
+			HTTPHost: string(measurement.Input),
+		},
+		Session: sess,
+		Target:  fmt.Sprintf(m.config.TestHelperAddress),
+	}
+	tk, err := g.Get(ctx)
+	measurement.TestKeys = tk
```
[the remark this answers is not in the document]
62w71st: Actually, the existing syntax was correct and consistent with other experiments. 

measurement.TestKeys is a struct which doesn't have TestKeys inside of it.

experiment/httphost/httphost.go
62w71st: Done.

62w71st:
Tests must pass now. PTAL.

62w71st:
@bassosimone Simone, I fixed up the commits and retested with `./miniooni http_host_header `

experiment/httphostheader/httphostheader.go
```diff
@@ -0,0 +1,78 @@
...
+		Begin: measurement.MeasurementStartTimeSaved,
+		Config: urlgetter.Config{
+			HTTPHost: string(measurement.Input),
+		},
+		Session: sess,
+		Target:  fmt.Sprintf(m.config.TestHelperURL),
+	}
+	tk, _ := g.Get(ctx)
+	measurement.TestKeys = tk
+	return nil
```
bassosimone, line 72: Here I've changed the code to return `nil` rather than `g.Get`'s error. Cases of censorship where the connection is reset cause `g.Get` to return an error. But the caller may not submit the measurement if we return `err` here. Thus, always returning `nil` is a much better choice here.

This semantics was not documented. So, as part of the same commit, I have also ensured that it's documented otherwise contributors, who cannot read my mind, are not aware of this fact 🙈.

(This semantics may not be the best one, and it may make sense to always submit the result of a measurement. I should probably open an issue about this to spend some more time thinking about this topic and then, possibly, bulk changing all the experiments to propagate errors back.)

bassosimone (approved):
Awesome! I am very happy to see this building block being ready! Thank you!

🐳 🎸

experiment/httphostheader/httphostheader.go
bassosimone: Issue here: https://github.com/ooni/probe-engine/issues/1017

PR #965, opened 2020-10-05
refactor(experiment.go): extract ExperimentBuilder and list of experiments

bassosimone:
This change extracts from experiment.go the ExperimentBuilder struct
as well as the list of all the experiments. It makes working with
experiment.go more manageable by moving separate components into two
distinct separate files. This change is yak shaving while I am
investigating https://github.com/ooni/explorer/issues/495.


bassosimone:
Merging with a failure occurring in an unrelated component.

PR #967, opened 2020-10-05
refactor(generate_*): better names for tests

bassosimone:
This is refactoring after we have addressed the main issue that
was causing https://github.com/ooni/explorer/issues/495.


bassosimone:
Ah, oops, this is not possible because it leads to duplicate names.

PR #968, opened 2020-10-05
fix(model): make sure we honour sharing settings

bassosimone:
With https://github.com/ooni/probe-engine/pull/966 and this change
together we have a more in depth mechanism for enforcing that we're
doing the right thing in terms of honouring sharing settings.

Yet, my instinct is that we should push for removing this vestigial
options that increase complexity and reduce the quality of data we
receive, in the mid term. Shall discuss this with the rest of the team.

### DIFF
--- a/model/model_test.go
+++ b/model/model_test.go
@@ -58,15 +58,29 @@ func TestAddAnnotations(t *testing.T) {
 	}
 }
 
-func makeMeasurement(probeIP, probeASN, probeCC string) model.Measurement {
+type makeMeasurementConfig struct {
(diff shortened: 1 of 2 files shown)


bassosimone:
This is related to https://github.com/ooni/explorer/issues/495. Not strictly required for fixing the original issue but I felt like it was a good idea to check around for other parts in the code base and have more consistency and robustness.

bassosimone:
We can merge with the usual intermitted fbmessenger failure: https://github.com/ooni/probe-engine/issues/906.

PR #970, opened 2020-10-06
miniooni: sync -g option with other probes

bassosimone:
See https://github.com/ooni/explorer/issues/495#issuecomment-704101604

While there, add a QA script for making sure everything is WAI.

Interestingly this led to discover a minor issue where the network
name was not included when empty.


bassosimone:
I am going to merge while the golang / test check is running since the golang / pull_request already completed ✅

PR #979, opened 2020-10-13
dns: set dot to ALPN

### LINKED ISSUE #657: dns: set ALPN to dot when doing DoT
See https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids

Spotted by @fortuna

This boils down to populating a TLS certificate and making sure the ALPN is correct at https://github.com/ooni/probe-engine/blob/master/netx/httptransport/httptransport.go#L269

xhdix:
Fix #657
`$ ./miniooni -i dot://1.0.0.1 dnscheck` :
image

### DIFF
--- a/netx/netx.go
+++ b/netx/netx.go
@@ -277,9 +277,7 @@ func NewDNSClientWithOverrides(config Config, URL, hostOverride, SNIOverride str
 	if err != nil {
 		return c, err
 	}
-	if SNIOverride != "" {
-		config.TLSConfig = &tls.Config{ServerName: SNIOverride}
-	}
+	config.TLSConfig = &tls.Config{ServerName: SNIOverride}
 	switch resolverURL.Scheme {
 	case "system":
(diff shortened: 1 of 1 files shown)


netx/netx.go
```diff
@@ -308,6 +306,7 @@ func NewDNSClientWithOverrides(config Config, URL, hostOverride, SNIOverride str
 		c.Resolver = resolver.NewSerialResolver(txp)
 		return c, nil
 	case "dot":
+		config.TLSConfig = &tls.Config{NextProtos: []string{"dot"}}
```
bassosimone, line 309:
```suggestion
		config.TLSConfig.NextProtos = []string{"dot"}
```

We don't want to create a new config here. If we do, we loose the SNI override above.

xhdix: Thanks. done!

bassosimone:
Thanks!

PR #982, opened 2020-10-13
fix(libminiooni): more robust file-opening code

### LINKED ISSUE #980: Read from file error: New line difference in Unix(LF) and Windows(CRLF)
**To Reproduce**
Create a text file in Windows and use it in Linux!

```
[     10.806731] <info> [24/32] running with input: https://dns.google/dns-query
[     10.806815] <warn> measurement failed: map[error:the input URL is invalid: parse "https://dns.google/dns-query\r": net/url: invalid control character in URL]
[     10.806826] <info> submitting measurement to OONI collector; please be patient...
[     10.806856] <debug> httpx: request body: 920 bytes
[     10.806863] <debug> httpx: method: POST
[     10.806869] <debug> httpx: URL: https://ps2.ooni.io/report/20201013T132548Z_AS58224_CMHxqiDagokNTdbDPAF9Svp7cXSSHxvtKJEixKXhDtvS6DJFLi
```

xhdix:
Fix #980


libminiooni/libminiooni.go
```diff
@@ -235,14 +235,24 @@ func loadFileInputs(opts *Options) {
...
 		}
-		content, err := ioutil.ReadFile(opts.InputFilePath)
+		info, err := os.Stat(opts.InputFilePath)
+		if os.IsNotExist(err) {
+			fatalOnError(err, "input file does not exist")
+		}
+		if info.IsDir() {
+			fatalWithString("input file is a directory")
+		}
+		file, err := os.Open(opts.InputFilePath)
```
bassosimone, line 245: We don't want to be this precise here and only here. There are many places in the codebase where we do not check whether this specific error condition occurs and I don't think they should be neglected.

The code you have written here should be put into a convenience function inside another package. I believe that, for example, it would make sense to create an `internal/fsx` package with file system extensions. I am not suggesting to use `osx` because it would be confusing. I also know Go is planning on introducing `fs` in the standard library that will abstract over the filesystem, hence the name I am proposing.

I think it's fine to introduce this new functionality into the current PR, to do less bureaucratic work, as long as we also mention this new functionality into the title of the PR. (While there, I think we already deal with EOL in the input file: what we're doing here is perhaps fix the way in which we do that?)

I would also be grateful if you could write a test for the new functionality (the name could be `Open`) where we verify that an error is returned in all the cases we care about. For this purpose, we can use a `testdata` directory inside `interal/fsx` where the proper structure of files and directories is pre-added and committed. (We can use a `.gitignore` file containing `*` to force git to keep an otherwise empty directory.)

When these changes are applied, then:

1. we can create an issue about fixing all the other places in which we `os.Open` a file

2. we can apply the suggested diff here and simplify the codebase

```suggestion
		file, err := fsx.Open(opts.InputFilePath)
```

libminiooni/libminiooni_unit_test.go
```diff
@@ -26,12 +26,29 @@ func TestFileInputInvalidFile(t *testing.T) {
...
 	defer func() { recover() }()
 
+	dir += "/test.txt"
 	opts := Options{InputFilePath: dir}
 	loadFileInputs(&opts)
 
 	t.Errorf("expected panic with invalid input file")
 }
 
+func TestFileInputDirectory(t *testing.T) {
```
bassosimone, line 36: Properly adapted, this could become a test for `fsx.Open`. (To be clear, `fsx.Open` should return `*os.File, error` rather than panicking as it does here, since it's a convenience/library function.)

bassosimone (changes requested):
Thanks for spotting this interesting issue. I have suggested a way forward where we make sure your changes could also be applied to other places in the codebase with this issue.

🎸

libminiooni/libminiooni.go
xhdix: I agree.

> 1. we can create an issue about fixing all the other places in which we os.Open a file

Other places are to save the file, not to open it. So I also should write a code to save the file?(To have a complete `fsx`?)

xhdix: (Also: I think I'm not good at naming. :smile:  )

internal/fsx/fsx.go
```diff
@@ -0,0 +1,19 @@
...
+	"fmt"
+	"os"
+)
+
+func Open(name string) (*os.File, error) {
+	file, err := os.Open(name)
+	if err != nil {
+		return nil, err
+	}
+	info, err := file.Stat()
```
bassosimone, line 14:
```suggestion
	info, err := file.Stat()
	if err != nil {
		return nil, err
	}
```

We want to check the return value from stat before proceeding.

internal/fsx/fsx.go
```diff
@@ -0,0 +1,19 @@
...
+	if info.IsDir() {
+		return nil, fmt.Errorf("input file is a directory: " + name)
+	}
+	return file, err
```
bassosimone, line 18:
```suggestion
	return file, nil
```

Let us just return `nil` here, since we're morally returning a sum type.

internal/fsx/fsx_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"bufio"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/ooni/probe-engine/internal/fsx"
+)
+
+var StateBaseDir = "../../testdata/fsx/"
```
bassosimone, line 12:
```suggestion
var StateBaseDir = "./testdata/fsx/"
```

I suggest to create this directory as `internal/fsx/testdata`, so we have less cross references between different parts of the codebase, which is in general a very nice property to have.

testdata/.gitignore
```diff
@@ -8,3 +8,4 @@
 /psiphon_config.json
 /psiphon_unit_tests/
 /test-download-resources-idempotent*/
+/fsx/
```
bassosimone, line 11: This directory should not be ignored. On the contrary, we want to create files and directories inside of `./internal/fsx/testdata` so that we don't need to create them during tests (see my above comments).

internal/fsx/fsx_test.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	defer os.Remove(inputfile.Name())
+
+	if _, err := inputfile.Write(file_contents); err != nil {
+		t.Fatal(err)
+	}
+	if err := inputfile.Close(); err != nil {
+		t.Fatal(err)
+	}
+
```
bassosimone, line 64:
```suggestion
```

Please create a file in `./internal/fsx/testdata` and commit it. So we don't need to code above and we can just open such file in our test, resulting in a shorter and easier to understand test implementation.

internal/fsx/fsx_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func TestFileInputDirectory(t *testing.T) {
+	dir, err := ioutil.TempDir(StateBaseDir, "test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
```
bassosimone, line 38: Please just open an existing directory (for example `./internal/fsx/testdata`

internal/fsx/fsx_test.go
```diff
@@ -0,0 +1,74 @@
...
+	// test function panics with invalid directory
+	defer func() { recover() }()
+
+	_, err = fsx.Open(dir)
+	if err == nil {
+		t.Errorf("expected panic with input directory")
```
bassosimone, line 44:
```suggestion
		t.Fatalf("not the error we expected: %+v", err)
```

I think we also need to update the error string here.

internal/fsx/fsx_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/ooni/probe-engine/internal/fsx"
+)
+
+var StateBaseDir = "../../testdata/fsx/"
+
+func TestFileInputInvalidFile(t *testing.T) {
```
bassosimone, line 14:
```suggestion
func TestOpenNonexistentFile(t *testing.T) {
```

internal/fsx/fsx_test.go
```diff
@@ -0,0 +1,74 @@
...
+	dir, err := ioutil.TempDir(StateBaseDir, "test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
```
bassosimone, line 20: Just use a nonexisting file path inside `./internal/fsx/testdata`, please.

internal/fsx/fsx_test.go
```diff
@@ -0,0 +1,74 @@
...
+	// test function panics with invalid file
+	defer func() { recover() }()
+
+	dir += "/test.txt"
+	_, err = fsx.Open(dir)
+	if err == nil {
```
bassosimone, line 26: I think we should made this check more robust 🤔

bassosimone (changes requested):
Thanks a bunch. I have done a thorough review of this diff and suggested small style changes to make this code more in line with the rest of the tree. There is a single place where I have not been able to make my mind regarding how to properly check an error in the tests.

🎸

bassosimone (approved):
Thank you so much for helping to improve ooni/probe-engine!

I am going to merge you PR with small editorial changes once I figure out what is wrong with the test or once I have confidence everything is working locally, whichever happens first.

(Update: it seems the main issue with the test suite is https://github.com/ooni/probe-engine/issues/1005.)

🐳

bassosimone:
Solved conflicts manually and merged here: https://github.com/ooni/probe-engine/pull/1013

bassosimone:
Thank you again!

Issue #1005, opened 2020-10-27
Investigate failed build with data races


bassosimone:
See https://github.com/ooni/probe-engine/runs/1315346306. It seems the main source of failures is that power DNS did not look super happy about us using it. At the same time, there was also a data race caused by a failure.

bassosimone:
Similar failure: https://github.com/ooni/probe-engine/pull/1006/checks?check_run_id=1316245193. This seems now systematic. I am wondering whether doh.powerdns.org is unhappy with our tests run from GHA?

bassosimone:
This is actually priority/high now. Tests fail also locally. It seems powerdns has implemented rate limiting.

bassosimone:
The first step towards fixing this issue has been https://github.com/ooni/probe-engine/pull/1007. We have refactored the tests such that most tests run daily or weekly. We have disabled many integration tests, resulting in less coverage, quicker builds, and, critically, a green tree because the failing tests were among the disabled tests. Now that we have a green tree, we can then refine this work and also understand why those integration tests are failing and how to remediate. We also need to take care of the data race. Luckily, I can reproduce every issue also locally.

bassosimone:
In https://github.com/ooni/probe-engine/pull/1008, I've further increased the time required to run the whole test suite by starting to flag "long" test as such and by running by default only the "short" tests. This seems a better distinction between unit and integration tests, meaning that we probably want to run for every commit only the tests that are quick.

bassosimone:
With https://github.com/ooni/probe-engine/pull/1009 I am now satisfied of the quick tests. Now it's "just" a matter of re-adding more of the tests that were previously removed so they run daily.

bassosimone:
In https://github.com/ooni/probe-engine/pull/1010, I have added code to run all the tests every day. Now, it may probably worth thinking whether we want to use the "short" technique also for all the previously disabled tests.

bassosimone:
In https://github.com/ooni/probe-engine/pull/1011 I have moved from tests from unit to integration and I have also improved the documentation around what tests are flagged as integration tests. Using tags is probably useful compared to having to manually flag every long test as such? I have not fully wrapped my mind around what's better.

bassosimone:
With the merging of https://github.com/ooni/probe-engine/pull/1012, we now have fully mitigated the original problem. We have also reorganised and rationalised our testing, which is also a very good thing.

bassosimone:
[...]

PR #1006, opened 2020-10-27
fix: don't use snake case for variable names

bassosimone:


bassosimone:
Failures similar to the ones documented at https://github.com/ooni/probe-engine/issues/1005. I am convinced that this issue is caused by doh.powerdns.org failing for every request we issue from GHA now. Maybe some rate limiting.

Because I am convinced of the above, I think it's safe to merge this innocuous fix with failing tests.

PR #1012, opened 2020-10-29
fix(sessionresolver): use dns.quad9.net

bassosimone:
See https://github.com/ooni/probe-engine/issues/1005


bassosimone:
So, we see that with b681760 all tests are now green. This means that the real issue in https://github.com/ooni/probe-engine/issues/1005 was that doh.powerdns.org had stopped working. We can now merge this PR.

Issue #1018, opened 2020-10-30
netx: make netx.CertPool private


bassosimone:
Noticed it being public may be confusing during https://github.com/ooni/probe-engine/pull/978 review.

cyBerta:
Currently it seems to be the only way to set different CA certs. If the CertPool becomes private, it would be cool to be able to assign either certs (e.g. array of pem string formatted entries) or a configured certpool to a urlGetter.Config.

bassosimone:
> If the CertPool becomes private, it would be cool to be able to assign either certs (e.g. array of pem string formatted entries) or a configured certpool to a urlGetter.Config.

Yes, we need to implement this to allow for your use case (see https://github.com/ooni/probe-engine/issues/1019). Will look into it on Monday! Thank you!

bassosimone:
Done in https://github.com/ooni/probe-engine/pull/1020

Issue #1019, opened 2020-10-31
urlgetter, netx: allow to add extra certificates


bassosimone:
As highlighted by @cyBerta in https://github.com/ooni/probe-engine/issues/1018#issuecomment-719922725, we need to support the use case of adding extra certificates to urlgetter and netx. This is quite urgent since it's blocking the implementation of the riseupvpn experiment in https://github.com/ooni/probe-engine/pull/978.

bassosimone:
Done in https://github.com/ooni/probe-engine/pull/1020!

PR #1020, opened 2020-11-03
fix(netx, urlgetter): allow to use arbitrary cert pools

bassosimone:
See https://github.com/ooni/probe-engine/issues/1019


bassosimone:
@cyBerta could you please review this PR and let me know whether you think it's good for your use case? Thanks! 🙏

cyBerta:
works perfectly!
Also it seems this closes #1018

bassosimone:
@cyBerta thanks for looking into this PR! I'm going to merge once my next meeting is over. Regarding the conflict, I've fixed it in a way that makes it further obvious that there were just additions into the `Config` structure.

Issue #866, opened 2020-08-20
go1.15: go build -v fails unless -tags PSIPHON_DISABLE_QUIC is passed


bassosimone:
(We have now [completely mitigated](https://github.com/ooni/probe-engine/issues/866#issuecomment-696170540) this issue by building atop the [documented workaround](https://github.com/ooni/probe-engine/issues/866#issuecomment-689532954).)

This is the output I obtain when compiling using Go 1.15:

```
% ./miniooni -h
panic: qtls.ConnectionState not compatible with tls.ConnectionState

goroutine 1 [running]:
github.com/Psiphon-Labs/quic-go/internal/handshake.init.1()
	/Users/sbs/go/pkg/mod/github.com/!psiphon-!labs/quic-go@v0.14.1-0.20200306193310-474e74c89fab/internal/handshake/unsafe.go:17 +0x12e
```

The reason seems to be that Go 1.15 changed its `tls.ConnectionState` structure. There is [a fix applied upstream](https://github.com/lucas-clemente/quic-go/commit/125318d9c948c380e1a8d3421bdbcada437efe0a) that handles this issue. It seems Psiphon is using its own fork of quic-go: https://github.com/Psiphon-Labs/quic-go.

This means that, for fixing the issue, the Psiphon devs need to backport the patch. For now, it's important to use Go 1.14 only for building OONI.

bassosimone:
## Documented workaround

Psiphon devs just recommended this workaround: `go build -a -v -tags "DISABLE_QUIC"` which has been confirmed to be working for them when using Psiphon and Go 1.15. Many thanks!

cyBerta:
I adapted the build script accordingly. Is it worth a PR?

bassosimone:
Who's building using `go build -v ./...` is gonna get a broken binary anyway, _unless_ the proper tag is passed from command line. Because of this reasoning, I believe fixing the script to _conditionally_ disable QUIC when developing using Go 1.15 is not fully solving the problem, because we're producing a broken binary for who's not using the build script.

How about: we create one/two files in the top-level package that only conditionally build if either we're in Go 1.14 or the user has provided the `DISABLE_QUIC` tag? With this solution, the build _breaks_. This is arguably better than breaking the binary, which is something that is definitely going to be much more surprising for the user?

What do you think?

cyBerta:
[...]

PR #1022, opened 2020-11-04
fix: allow building w/o -tags PSIPHON_DISABLE_QUIC

bassosimone:
Releasing a version of probe-engine that only builds with `-tags PSIPHON_DISABLE_QUIC` for every version of Go is a bad idea. This PR is based of https://github.com/ooni/psiphon, which is vendors psiphon sources and all its dependencies in a way that does not cause conflicts in the package path.

This codebase also disables QUIC support, so we don't have any conflict and the code builds seamlessly when using Go 1.15 (now the default on, e.g., Fedora). Because I've been told by Psiphon developers that they're working to upgrade their QUIC implementation, this means we are not going to use Psiphon without QUIC for a long time. Also, they told me it should ~fine to use Psiphon without QUIC anyway.

With this diff committed, we dramatically simplify dependency management and we remove the requirement of specifying a manual flag to get probe-engine (and hence probe-cli) to compile with Go 1.15.

This is the reference issue: https://github.com/ooni/probe-engine/issues/866.


bassosimone:
What is missing as of https://github.com/ooni/probe-engine/pull/1022/commits/78e932be3e065dba041e968739866fd599a95ca1:

- [ ] engine: update instructions regarding upgrading dependencies
- [ ] cli: update instructions regarding upgrading dependencies
- [ ] anything else?

PR #1029, opened 2020-11-08
dns dialer for http3 transport

kelmenhorst:


netx/httptransport/http3transport_test.go
```diff
@@ -2,18 +2,25 @@ package httptransport_test
...
 )
 
+type MockHTTP3Dialer struct{}
+
+func (d MockHTTP3Dialer) Dial(network, host string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error) {
+	return quic.DialAddrEarly(host, tlsCfg, cfg)
+}
 func TestUnitHTTP3TransportSuccess(t *testing.T) {
-	txp := httptransport.NewHTTP3Transport(selfcensor.SystemDialer{}, dialer.TLSDialer{})
+	txp := httptransport.NewHTTP3Transport(httptransport.Config{selfcensor.SystemDialer{}, MockHTTP3Dialer{}, dialer.TLSDialer{}})
```
bassosimone, line 23: Could you explicitly initialise fields? I think this makes the code more obvious.

(I.e., `X{A: a, B: b}` rather then `X{a, b}`)

The motivation of this request is that adding explicit initialisation of fields allows us to take the liberty of adding new fields and changing the order of fields in the `Config` structure w/o breaking the build.

netx/httptransport/http3transport_test.go
```diff
@@ -32,7 +39,7 @@ func TestUnitHTTP3TransportSuccess(t *testing.T) {
 }
 
 func TestUnitHTTP3TransportFailure(t *testing.T) {
-	txp := httptransport.NewHTTP3Transport(selfcensor.SystemDialer{}, dialer.TLSDialer{})
+	txp := httptransport.NewHTTP3Transport(httptransport.Config{selfcensor.SystemDialer{}, MockHTTP3Dialer{}, dialer.TLSDialer{}})
```
bassosimone, line 42: Could you explicitly initialise fields? I think this makes the code more obvious.

netx/httptransport/httptransport.go
```diff
@@ -7,6 +7,13 @@ import (
 	"net/http"
 )
 
+// Config contains the configuration assumed by this package
```
bassosimone, line 10:
```suggestion
// Config contains the configuration required for constructing an HTTP transport
```

This seems more explicit?

netx/netx.go
```diff
@@ -168,6 +175,15 @@ func NewDialer(config Config) Dialer {
 	return d
 }
 
+// NewHTTP3DNSDialer creates a new DNS Dialer for HTTP3 transport, with the resolver from the specified config
+func NewHTTP3DNSDialer(config Config) HTTP3Dialer {
```
bassosimone, line 179: I would call this function `NewHTTP3Dialer` since the returned type is an `HTTP3Dialer`

netx/netx.go
```diff
@@ -234,14 +253,13 @@ func NewHTTPTransport(config Config) HTTPRoundTripper {
...
 type httpTransportInfo struct {
-	Factory       func(httptransport.Dialer, httptransport.TLSDialer) httptransport.RoundTripper
+	Factory       func(httptransport.Config) httptransport.RoundTripper
 	TransportName string
 }
 
 var allTransportsInfo = map[bool]httpTransportInfo{
 	false: httpTransportInfo{
-		Factory: httptransport.NewSystemTransport,
-		// TODO(kelmenhorst): distinguish between h2 / http/1.1
```
bassosimone, line 244: Please, do not remove this comment.

netx/dialer/http3dns.go
```diff
@@ -0,0 +1,56 @@
...
+			return nil, err
+		}
+		sess, err := quic.DialEarly(udpConn, udpAddr, host, tlsCfg, cfg)
+		if err == nil {
+			return sess, nil
+		}
+		errorslist = append(errorslist, err)
+	}
+	return nil, reduceErrors(errorslist)
+}
```
bassosimone, line 48: We should write some lightweight integration test that tests this part of the codebase and lives inside the `dialer` package. To this end we can adapt perhaps some more code from `dns_test.go`.

bassosimone (changes requested):
Very good work! I just tested it locally and it's working as intended! I am also pleased by the shape the QUIC code is taking and by the swift progress we're making.

To have this code merged, we mainly need to write some more tests in `http3dns.go`. Apart from that, I have highlighted a couple of style changes I'd like to see applied.

Thank you! 🎸

netx/dialer/http3dns.go
```diff
@@ -0,0 +1,65 @@
...
+
+// LookupHost implements Resolver.LookupHost
+func (d HTTP3DNSDialer) LookupHost(hostname string) ([]string, error) {
+	if net.ParseIP(hostname) != nil {
+		return []string{hostname}, nil
+	}
+	return d.Resolver.LookupHost(context.Background(), hostname)
```
bassosimone, line 64:
```suggestion
	// TODO(bassosimone,kelmenhorst): here we should actually use the context
	// passed to dial when we upgrade to a version of quic-go that allows us to do so.
	return d.Resolver.LookupHost(context.Background(), hostname)
```

netx/dialer/http3dns_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	dialer := dialer.HTTP3DNSDialer{Resolver: new(net.Resolver)}
+	sess, err := dialer.Dial("udp", "www.google.com:443", tlsConf, &quic.Config{})
+	if err != nil {
+		t.Fatal("unexpected error")
+	}
+	if sess == nil {
+		t.Fatal("non nil sess expected")
+	}
+}
```
bassosimone, line 27:
```suggestion
}

```

netx/httptransport/http3transport_test.go
```diff
@@ -2,18 +2,25 @@ package httptransport_test
...
 	"github.com/ooni/probe-engine/netx/dialer"
 	"github.com/ooni/probe-engine/netx/httptransport"
 	"github.com/ooni/probe-engine/netx/selfcensor"
 )
 
+type MockHTTP3Dialer struct{}
+
+func (d MockHTTP3Dialer) Dial(network, host string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error) {
+	return quic.DialAddrEarly(host, tlsCfg, cfg)
+}
```
bassosimone, line 21:
```suggestion
}

```

bassosimone (commented):
I've tested it locally and it was working as intended! I modified a test such that it works not only on Linux but also on macOS! Thank you so much, merging now! 🐳

bassosimone (approved):

Issue #994, opened 2020-10-23
ci failure: context timeout not working as intended


bassosimone:
Happens periodically. Error like:

```
2020-10-23T12:51:14.5182055Z === RUN   TestNewContextWithNegativeTimeout
2020-10-23T12:51:14.5183938Z ##[error]    sessioncontext_test.go:54: context timeout not working as intended
```

See https://github.com/ooni/probe-engine/runs/1298002926

bassosimone:
Here's another one: https://github.com/ooni/probe-engine/runs/1330577114

bassosimone:
Here's another one: https://github.com/ooni/probe-engine/pull/1030/checks?check_run_id=1378817750

bassosimone:
Here's another one: https://github.com/ooni/probe-engine/pull/1042/checks?check_run_id=1394298622

PR #1030, opened 2020-11-10
chore: start using ams-pg-test.ooni.org for testing

bassosimone:
See https://github.com/ooni/backend/issues/461


bassosimone:
I am going to merge since the only failure above is an instance of https://github.com/ooni/probe-engine/issues/994.

Issue #1026, opened 2020-11-05
fix(geolocate): implement fallback for lookup IP


bassosimone:
We noticed cases where the IP lookup failed and hence the measurement failed. It seems we can increase the robustness of the measurement by chaining the IP resolvers such that we try more than one of them in a loop.

bassosimone:
One initial problem I noticed with the geolookup code was that it was discarding useful errors. So in https://github.com/ooni/probe-engine/pull/1032 I have added code that ensures we return to the caller information on all errors that occurred.

bassosimone:
In https://github.com/ooni/probe-engine/pull/1033 I've started adding more options.

bassosimone:
We probably need to add another rule to the code: if there is no IP address then we return an error. I am now starting to wonder whether this is what's causing https://github.com/ooni/probe-engine/issues/988.

PR #1032, opened 2020-11-10
fix(geolocate): do not hide underlying errors

bassosimone:
See https://github.com/ooni/probe-engine/issues/1026

### DIFF
--- a/internal/multierror/multierror.go
+++ b/internal/multierror/multierror.go
@@ -1,0 +1,66 @@
+// Package multierror contains code to manage multiple errors.
+package multierror
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// Union is the logical union of several errors. The Union will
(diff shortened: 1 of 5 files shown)


internal/multierror/multierror.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// AddWithPrefix adds the specified child error to the Union error
+// with the specified prefix before the child error.
+func (err *Union) AddWithPrefix(prefix string, child error) {
+	err.Add(fmt.Errorf("%s: %w", prefix, child))
+}
+
+// Is returns whether the Union error contains at least one child
+// error that is exactly the specified target error.
```
bassosimone, line 41:
```suggestion
// Is returns whether the Union error is the specified target error or whether it
// contains at least one child error that is exactly the specified target error.
```

Issue #974, opened 2020-10-09
engine, cli: remove include_{asn,cc,ip} settings


bassosimone:
No point in having these settings. They only create complexity and, in any case, what we mostly want is to have the CC and the ASN. We don't want the IP address, so why asking for it? We cannot do much when there is not ASN, and this actually leads to rather frustrating measurements, so why asking people whether they want to opt out? As for the country, all the deployed apps already don't give users the option to opt out of that, so why bothering with a setting inside the engine?

Stems from https://github.com/ooni/explorer/issues/495#issuecomment-704101604.

bassosimone:
Pull request https://github.com/ooni/probe-engine/pull/1063 removes the privacy settings and adapts the tests such that both short and long tests continue to pass. The approach in such a PR has been that of not making any radical or refactoring change and of leaving there a bunch of TODO entries. The following list enumerates such TODOs:

- [ ] [experiment_generated_test.go: refactor/rename](https://github.com/ooni/probe-engine/pull/1063/files#diff-8304b64b1a6cdc8cfb74adf422e64719be89c89511c36ea194be35803cdff843R9)
- [ ] [libminiooni/libminiooni.go: remove unnecessary options (certainly -g and possibly others)](https://github.com/ooni/probe-engine/pull/1063/files#diff-dec40774a4b36a4ee77050acc27850a5ccf95d3ad5e7c55a3dd254bacd21c30fR40)
- [ ] [model/model_test.go: improve implementation quality](https://github.com/ooni/probe-engine/pull/1063/files#diff-261992f780b0563aea8800f78dd6d39466511838b2fda8becc9a0746efdd5d19R144)
- [ ] [model/privacysettings.go: refactor/rename](https://github.com/ooni/probe-engine/pull/1063/files#diff-f59905862965551dc309e72a692f23d20977fbbee98efa1108ee88364eb16728R10)
- [ ] [oonimkall/task_integration_test.go: do we need to update other tests in oonimkall/...](https://github.com/ooni/probe-engine/pull/1063/files#diff-9b666fcef8cb12368fc2d525f860d78d49702c39503cd99fdd6d6f042fb72c13R528)
- [ ] [oonimkall/task_integration_test.go: simplify TestPrivacyAndScrubbing](https://github.com/ooni/probe-engine/pull/1063/files#diff-9b666fcef8cb12368fc2d525f860d78d49702c39503cd99fdd6d6f042fb72c13R535)
[...]

PR #1063, opened 2020-11-24
feat: remove privacy settings

bassosimone:
From now on, we will always share ASN and CC and we will never
share the user's IP. This is a sweet spot where we get enough data
to meaningfully analyse what's going. We seldom need to know the
user IP, therefore, let us just disable that setting.

Part of https://github.com/ooni/probe-engine/issues/974.

This PR also ensures all tests continue to pass. To this end, I've run
long tests locally on a Fedora 33 box.

We will follow up with other pull requests that further cleanup the
codebase after this one has landed. There is a bunch of cases in
which we could easily refactor and simplify the codebase here.


model/model_test.go
```diff
@@ -123,13 +118,13 @@ func TestScrubCommonCase(t *testing.T) {
...
-		t.Fatal("ProbeIP has not been scrubbed")
+		t.Fatal("ProbeIP HAS NOT been scrubbed")
 	}
 	if m.ProbeNetworkName != config.ProbeNetworkName {
 		t.Fatal("ProbeNetworkName has been scrubbed")
 	}
-	if m.ResolverIP == config.ResolverIP {
-		t.Fatal("ResolverIP has not been scrubbed")
+	if m.ResolverIP != config.ResolverIP {
+		t.Fatal("ResolveIP has been scrubbed")
```
bassosimone, line 127:
```suggestion
		t.Fatal("ResolverIP has been scrubbed")
```

Issue #1057, opened 2020-11-23
quic/h3: quality assurance (1/n)


bassosimone:
We have an h3 implementation in tree. We wanna run a first qa step where we compare h3 with what's currently possible to do with urlgetter (e.g. setting a custom SNI, using a custom CA bundle). We wanna come up with a list of things to improve in h3 such that we can claim to have reduced the gap between h3 and the h2 implementation. We wanna fix these issues.

This is just one step towards this goal. It's fine to just focus on a bunch of sub-issues if there are too many and little time. We will anyways do similar work in subsequent sprints, if needed, until we've reached feature parity.

bassosimone:
- [ ] log transport protocol information in http logging debug-mode-only functionality

bassosimone:
## We can use the DNS cache with h3

This is already working, see:

```
[      2.003006] <info> [1/1] running with input: https://dns.google
[      2.003578] <debug> > GET https://dns.google
[      2.006223] <debug> > Accept-Language: en-US;q=0.8,en;q=0.5
[      2.006238] <debug> > User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.111 Safari/537.36
[      2.006252] <debug> > Host: dns.google
[      2.006262] <debug> > Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
[      2.006271] <debug> >
[      2.006347] <debug> resolve dns.google...
[      2.006415] <debug> resolve dns.google... ([8.8.8.8], <nil>) in 2.953µs
[      2.490670] <debug> < 200
[      2.490725] <debug> < Server: scaffolding on HTTPServer2
[      2.490741] <debug> < Content-Security-Policy: object-src 'none';base-uri 'self';script-src 'nonce-2FnbainU3UWMwPQA/WZEXg==' 'strict-dynamic' 'report-sample' 'unsafe-eval' 'unsafe-inline' https: http:;report-uri https://csp.withgoogle.com/csp/honest_dns/1_0;frame-ancestors 'none'
[      2.490755] <debug> < X-Content-Type-Options: nosniff
[      2.490765] <debug> < Content-Type: text/html; charset=UTF-8
[      2.490777] <debug> < X-Xss-Protection: 0
[      2.490787] <debug> < X-Frame-Options: SAMEORIGIN
[...]

PR #1079, opened 2020-11-26
context for http3 dialer

kelmenhorst:


netx/dialer/http3dns_test.go
```diff
@@ -127,11 +128,11 @@ func TestHTTP3DNSDialerDialEarlyFails(t *testing.T) {
 		NextProtos: []string{"h3-29"},
 	}
 	expected := errors.New("mocked DialEarly error")
-	mockDialEarly := func(net.PacketConn, net.Addr, string, *tls.Config, *quic.Config) (quic.EarlySession, error) {
+	mockDialEarly := func(context.Context, net.PacketConn, net.Addr, string, *tls.Config, *quic.Config) (quic.EarlySession, error) {
```
bassosimone, line 131:
```suggestion
	mockDialEarly := func(context.Background(), net.PacketConn, net.Addr, string, *tls.Config, *quic.Config) (quic.EarlySession, error) {
```

I think we should pass in a background context here?

netx/httptransport/http3transport.go
```diff
@@ -13,6 +14,21 @@ type HTTP3Dialer interface {
...
+// HTTP3WrapperDialer is a HTTP3Dialer that wraps a ContextDialer
+type HTTP3WrapperDialer struct {
+	Dialer HTTP3ContextDialer
+}
+
+// Dial implements HTTP3Dialer.Dial
+func (d HTTP3WrapperDialer) Dial(network, host string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error) {
+	return d.Dialer.DialContext(context.Background(), network, host, tlsCfg, cfg)
+}
+
```
bassosimone, line 31: I believe we should take a simpler approach here: 

```Go
// HTTP3Dialer is the definition of dialer for HTTP3 transport assumed by this package.
type HTTP3Dialer interface {
	DialContext(ctx context.Context, network, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error)
}
```

and then modify all the places we were using Dial to call DialContext with the proper context instead.

I believe you introduced the wrapper to preserve backwards compatibility with existing code that was already using `HTTP3Dialer`. Because we don't make API stability guarantees in probe-engine, I would recommend to just refactor existing users of `HTTP3Dialer` so we pass the context as the first argument. This allows us to reduce the overall [optional complexity](http://www.catb.org/~esr/writings/taoup/html/ch13s01.html#id2961759) of the codebase, which is ~always a good idea.

bassosimone (changes requested):
Thanks! This is a very good start and I am happy to see we're now able to pass the context around also for QUIC 🎸. I've provided suggestions on how to improve this diff.

netx/httptransport/http3transport.go
kelmenhorst: There is actually no `DialContext` API for the http3.Roundtripper, you can only assign a custom `Dial` method when creating a new HTTP3 Transport. That's why I built this Wrapper.

bassosimone: Ah, that's a bit surprising! I assumed they also had added support in `http3.RoundTripper`. Then, I think it's fine to keep the wrapper around for the time being. (Maybe it makes sense to document why there's the wrapper?)

kelmenhorst: Yes, for sure, I will add an explanation.

netx/dialer/http3dns_test.go
kelmenhorst: I'm not exactly sure what you mean, mockDialEarly is called with context.Background() a few lines later, this is just the declaration. Or what am I missing here? :)

bassosimone: You're not missing anything, I think I've read the diff incorrectly :-)

bassosimone (approved):

bassosimone:
Merged, thank you!

bassosimone:
Xref: https://github.com/ooni/probe-engine/issues/1057

PR #1091, opened 2020-11-30
feat(engine): create Submitter abstraction

bassosimone:
This is a small abstraction that allows us to simplify code that is
submitting measurements by eliminating conditionals.

Work part of https://github.com/ooni/probe/issues/1283.

### DIFF
--- a/submitter.go
+++ b/submitter.go
@@ -1,0 +1,75 @@
+package engine
+
+import (
+	"context"
+
+	"github.com/ooni/probe-engine/model"
+)
+
+// TODO(bassosimone): maybe keep track of which measurements
+// could not be submitted by a specific submitter?
+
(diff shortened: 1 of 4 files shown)


submitter.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/ooni/probe-engine/model"
+)
+
+// TODO(bassosimone): maybe keep track of which measurements
+// could not be submitted by a specific submitter?
+
+// Submitter submits a measurement to the OONI collector.
+type Submitter interface {
+	// SubmitAndUpdateMeasurementContext submits the measurement
+	// and updates its report ID field in case of sucess.
```
bassosimone, line 15:
```suggestion
	// and updates its report ID field in case of success.
```

PR #1124, opened 2020-12-08
fix(urlgetter): progress over multiple collection tasks

cyBerta:
Currently the `Multi.Collect()` method calculates the progress of a task based on the length of the `MultiInput` array. If an experiment performs multiple times `Multi.Collect()` the progress repeats increasing from 0% to 100%. This is especially an issue for a GUI client such as OONI Android, as it shows confusing estimations about the expected remaining runtime of a test. The shown expected time decreases multiple times to 0s and starts over again with a higher expected remaining time.

This PR introduces a method `Multi.CollectOverall()` that calculates the progress over multiple collection tasks. It is used in the RiseupVPN experiment. 

I would be interested to know if you see other/better options to archive the same goal.

### DIFF
--- a/experiment/riseupvpn/riseupvpn.go
+++ b/experiment/riseupvpn/riseupvpn.go
@@ -205,7 +205,7 @@ func (m Measurer) Run(ctx context.Context, sess model.ExperimentSession,
 	}
 	multi := urlgetter.Multi{Begin: measurement.MeasurementStartTimeSaved, Getter: m.Getter, Session: sess}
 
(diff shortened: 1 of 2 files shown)


bassosimone:
This seems very useful, thank you! I will take a look!

bassosimone:
I took an initial look at this code and this looks good to me. I'll take another look next Monday!

Thank you!

bassosimone (approved):
🐳 Thank you!

PR #1126, opened 2020-12-08
quic: TLS handshake saver, error classification

kelmenhorst:
Part of  https://github.com/ooni/probe-engine/issues/1057.


kelmenhorst:
Tests are still missing, I just wanted to get some feedback on the overall structure.

bassosimone:
👀

netx/dialer/quic_dialer.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// QUICDialer is the definition of dialer for QUIC assumed by this package.
+type QUICDialer interface {
+	Dial(network, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error)
+}
+
+// QUICSystemDialer is a base dialer for QUIC
+type QUICSystemDialer struct {
+	Saver *trace.Saver
```
bassosimone, line 32: I see you have a `QUICSaverDialer` below. So maybe it's possible to not have a saver here? It also feels more today this way, probably?

netx/dialer/quicsaver.go
```diff
@@ -0,0 +1,126 @@
...
+		Name:          "quic_handshake_start",
+		NoTLSVerify:   tlsCfg.InsecureSkipVerify,
+		TLSNextProtos: tlsCfg.NextProtos,
+		TLSServerName: tlsCfg.ServerName,
+		Time:          start,
+	})
+	sess, err := h.Dialer.DialContext(ctx, network, addr, host, tlsCfg, cfg)
+	stop := time.Now()
+
+	if sess == nil {
```
bassosimone, line 56: I feel like this is better written as

```suggestion
	if err != nil {
```

netx/dialer/quicsaver.go
```diff
@@ -0,0 +1,126 @@
...
+			TLSNextProtos: tlsCfg.NextProtos,
+			TLSServerName: tlsCfg.ServerName,
+			Time:          stop,
+		})
+		return sess, err
+	}
+	state := tls.ConnectionState{}
+	if sess != nil {
+		state = ConnectionState(sess)
+	}
```
bassosimone, line 71: I think we've already excluded the sess == nil case?

netx/dialer/quicsaver.go
```diff
@@ -0,0 +1,126 @@
...
+		h.Saver.Write(trace.Event{
+			Duration:      stop.Sub(start),
+			Err:           err,
+			Name:          "quic_handshake_done",
+			NoTLSVerify:   tlsCfg.InsecureSkipVerify,
+			TLSNextProtos: tlsCfg.NextProtos,
+			TLSServerName: tlsCfg.ServerName,
+			Time:          stop,
+		})
+		return sess, err
```
bassosimone, line 66: Should be

```suggestion
		return nil, err
```

(if possible)

netx/dialer/quicsaver.go
```diff
@@ -0,0 +1,126 @@
...
+		Duration: stop.Sub(start),
+		Err:      err,
+		NumBytes: count,
+		Name:     errorx.WriteOperation,
+		Time:     stop,
+	})
+	return count, err
+}
+
+func (c saverUDPConn) ReadMsgUDP(b, oob []byte) (n, oobn, flags int, addr *net.UDPAddr, err error) {
```
bassosimone, line 109: can we not name variables here and avoid the ugly underscore names below?

netx/dialer/quicsaver.go
```diff
@@ -0,0 +1,126 @@
...
+	start := time.Now()
+	_n, _oobn, _flags, _addr, err := c.UDPConn.ReadMsgUDP(b, oob)
+	var data []byte
+	if _n > -1 {
```
bassosimone, line 113: Why not `> 0` here?

netx/dialer/quicsaver.go
```diff
@@ -0,0 +1,126 @@
...
+type QUICHandshakeSaver struct {
+	Saver  *trace.Saver
+	Dialer QUICContextDialer
+}
+
+// DialContext implements QUICContextDialer.DialContext
+func (h QUICHandshakeSaver) DialContext(ctx context.Context, network string, addr string, host string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error) {
+	start := time.Now()
+	h.Saver.Write(trace.Event{
+		Name:          "quic_handshake_start",
```
bassosimone, line 47: Maybe you can define also this name in errorx?

netx/dialer/quicsaver.go
```diff
@@ -0,0 +1,126 @@
...
+		Time:          start,
+	})
+	sess, err := h.Dialer.DialContext(ctx, network, addr, host, tlsCfg, cfg)
+	stop := time.Now()
+
+	if sess == nil {
+		h.Saver.Write(trace.Event{
+			Duration:      stop.Sub(start),
+			Err:           err,
+			Name:          "quic_handshake_done",
```
bassosimone, line 60: Maybe you can define also this name in errorx?

netx/dialer/quicsaver_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+type MockQUICDialer struct {
+	Dialer dialer.QUICContextDialer
+	Sess   quic.EarlySession
+	Err    error
+}
+
+func (d MockQUICDialer) DialContext(ctx context.Context, network, addr string, host string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error) {
+	if d.Dialer != nil {
+		d.Dialer.DialContext(ctx, network, addr, host, tlsCfg, cfg)
```
bassosimone, line 26: I find it surprising that you don't `return` here.

netx/dialer/quicsaver_test.go
```diff
@@ -0,0 +1,207 @@
...
+	if ev[0].Time.After(time.Now()) {
+		t.Fatal("unexpected Time")
+	}
+	if ev[1].Duration <= 0 {
+		t.Fatal("unexpected Duration")
+	}
+	if ev[1].Err != nil {
+		t.Fatal("unexpected Err", ev[1].Err)
+	}
+	if ev[1].Name != "quic_handshake_done" {
```
bassosimone, line 162: Yeah, these strings maybe should just be defined once in errorx

netx/errorx/errorx.go
```diff
@@ -156,16 +163,23 @@ type SafeErrWrapperBuilder struct {
...
+	QuicErr bool
 }
 
 // MaybeBuild builds a new ErrWrapper, if b.Error is not nil, and returns
 // a nil error value, instead, if b.Error is nil.
 func (b SafeErrWrapperBuilder) MaybeBuild() (err error) {
+	failureString := toFailureString
+	if b.QuicErr {
+		failureString = toQUICFailureString
+	}
```
bassosimone, line 177: We should chat about this part of the design. I would have instinctively put checks for QUIC related errors into the function we're already using for checking for errors. Maybe I'm missing something?

netx/httptransport/http3transport.go
```diff
@@ -7,27 +7,18 @@ import (
...
-	Dial(network, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error)
-}
-
-// HTTP3ContextDialer is a dialer for HTTP3 transport using Context.
-type HTTP3ContextDialer interface {
-	DialContext(ctx context.Context, network, addr string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error)
-}
-
-// HTTP3WrapperDialer is a HTTP3Dialer that wraps a ContextDialer
+// QUICWrapperDialer is a QUICDialer that wraps a ContextDialer
```
bassosimone, line 13: (Maybe we can check whether perhaps they implemented this functionality?)

netx/netx.go
```diff
@@ -175,13 +175,19 @@ func NewDialer(config Config) Dialer {
...
-// NewHTTP3Dialer creates a new DNS Dialer for HTTP3 transport, with the resolver from the specified config
-func NewHTTP3Dialer(config Config) HTTP3Dialer {
+// NewQUICDialer creates a new DNS Dialer for QUIC, with the resolver from the specified config
+func NewQUICDialer(config Config) QUICDialer {
 	if config.FullResolver == nil {
 		config.FullResolver = NewResolver(config)
 	}
-	d := &dialer.HTTP3DNSDialer{Resolver: config.FullResolver}
-	var dialer HTTP3Dialer = &httptransport.HTTP3WrapperDialer{Dialer: d}
+	var d dialer.QUICContextDialer = &dialer.QUICSystemDialer{Saver: config.ReadWriteSaver}
```
bassosimone, line 183: I also mentioned above that I don't 100% understand this saver here. Maybe I'm missing something?

netx/dialer/quic_dialer.go
bassosimone: Ah, _maybe_ now I understand. This is the saver for the UDP connection while the other is a saver for dialing and for performing the QUIC handshake. I guess I'm a bit confused because I would have expected the saver dialer to take care of all these use cases, rather than having a bit of functionality here. Provided that I have understood correctly, I think it would be better to move this code into quicsaver.go. Otherwise, let's have a chat so that I can better understand what is the purpose of having Saver here!

bassosimone (commented):
Thank you so much! This PR contains terrific work!

My review:

1. provides many small style-related suggestions;

2. asks questions regarding the split between QUIC and normal errors;

3. asks whether it's possible to have all savers in the same place.

🎸

netx/netx.go
bassosimone: Update: I think now I understand its purpose. Please refer to this comment: https://github.com/ooni/probe-engine/pull/1126#discussion_r553470896.

netx/dialer/quicsaver_test.go
kelmenhorst: I did this in correspondence to _saver_test.go_ , seeing that, for instance, "tls_handshake_done" is also not defined in errorx.

netx/errorx/errorx.go
kelmenhorst: Yes, this would probably be more clear. In doing that, should, for example, a timeout error, induce the same notation in the result file, regardless whether it happened during a QUIC or a TCP connection? Or do you prefer having a `generic_timeout_error` and a `quic_generic_timeout_error`?

netx/dialer/quicsaver_test.go
bassosimone: Yes, I've just checked `saver.go`. It seems only the `tls_handshake_{start,done}` events do not have a corresponding name inside of `errorx.go`. I _think_ I understand what is going on. We have operations, whose name is defined in `errorx.go` and then we have events. Some events have the same name of operations, so we use the names in `errorx` while some others do not. This is the case of `"tls_handshake_done"`. It may be useful to have all these names in the same place to avoid typing errors, nonetheless. But maybe we can do this at a later time.

netx/errorx/errorx.go
bassosimone: I would keep a `generic_timeout_error` for now. I think maybe in the future we'll need to have more specific or different errors. But, for now, it seems to me we understand what is going on by combining operation and error.

(The error is `generic_timeout_error` rather than just `timeout` or `timeout_error` for legacy reasons that I honestly do not remember. Every time I write a table in LaTeX, I regret that this name is so long.)

netx/httptransport/http3transport.go
kelmenhorst: In the quic-go library, they only allow one function signature to be passed to the RoundTripper for Dialing, which is the one implemented by `QUICWrapperDialer` in this PR. This signature does not allow passing a context, which is why I did not consider it suitable to use for all of the QUIC dialers. The `QUICContextDialer` uses context.
And now that I'm looking into it, I realize that I don't even use the `QUICBaseDialer` anymore. :sweat_smile: Ok, so now we're down to 2 QUIC dialers.

netx/netx.go
kelmenhorst: Do you think the ReadWriteSaver should be encapsulated in a higher-level Dialer? In this case I'd have to provide access to the UDPconn struct to higher level Dialers.

netx/dialer/quic_dialer.go
```diff
@@ -0,0 +1,77 @@
...
+	ip := net.ParseIP(onlyhost)
+	udpAddr := &net.UDPAddr{IP: ip, Port: port, Zone: ""}
+	udpConn, err := net.ListenUDP("udp", &net.UDPAddr{IP: net.IPv4zero, Port: 0})
+	var pconn net.PacketConn = udpConn
+	if d.Saver != nil {
+		pconn = saverUDPConn{UDPConn: udpConn, saver: d.Saver}
+	}
+
+	sess, err := quic.DialEarlyContext(ctx, pconn, udpAddr, host, tlsCfg, cfg)
+	// TODO(kelmenhorst): Do we have to close udpConn ourselves? If so, when?
```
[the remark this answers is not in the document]
kelmenhorst: yes, this comment was outdated

netx/httptransport/http3transport.go
bassosimone: Too bad there are no party parrots on GitHub! 🎉

netx/netx.go
bassosimone: I'll ask you more details in chat. This part of the convo seems complex.

netx/dialer/quicsaver.go
kelmenhorst: right, but not with the comment above anymore :) unless sess == nil and err != nil are completely synonym here, which I'm not too confident about

bassosimone: This is where a language with a sum type would have helped. The semantics of this code is that we either have an error or the other thing is valid. It's used quite extensively in Go and also in probe-engine.

I agree that it would be much nicer to have `ErrorOr<Seession>` or `Error|Session` (even better), but sadly we're using a less advanced language where we humans need to do the sum type manually :^).

bassosimone: Still, if you want to enforce the sum type semantics here, you should intercept the case where error is nil and also session is nil and return yourself an error (there is code in the Go HTTP standard library that does that).

I think we should not write code this defensive though.

netx/netx.go
bassosimone: So, for the records. That is the only place where we can observe a UDP connection and wrap it. Doing this kind of wrapping elsewhere is possible, but seems also less practical.

netx/dialer/quic_dialer.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// DialContext implements QUICContextDialer.DialContext
+func (d QUICSystemDialer) DialContext(ctx context.Context, network string, addr string, host string, tlsCfg *tls.Config, cfg *quic.Config) (quic.EarlySession, error) {
+	onlyhost, onlyport, err := net.SplitHostPort(addr)
+	port, err := strconv.Atoi(onlyport)
+	if err != nil {
+		return nil, err
+	}
+	ip := net.ParseIP(onlyhost)
```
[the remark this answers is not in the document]
kelmenhorst: I added a test for ip == nil here.

netx/dialer/quicsaver_test.go
kelmenhorst: yes, if so, it should be changed for TCP and QUIC both, I think

netx/httptransport/http3transport.go
kelmenhorst: IKR? :D

netx/dialer/quicsaver_test.go
bassosimone: I agree with you. This does not seem a blocker to me. Let's do nothing for now. The code is already consistent as is.

netx/errorx/errorx.go
kelmenhorst: So I moved the QUIC specific content from the `toQUICFailureString` function into the `toFailureString` function and updated the tests. I added the QUIC specific failures in a batch at the end of the function. I'm not sure if the order should instead be based on the failure type.

netx/dialer/quic_dialer.go
```diff
@@ -0,0 +1,77 @@
+package dialer
```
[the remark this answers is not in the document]
kelmenhorst: So I added another package, `quicdialer`, that contains all of the implemented dialers and savers for QUIC.

bassosimone (approved):
I've read this diff very carefully, checked out the code myself, run tests, added formatting changes, and other minor changes to this pull request.

The code in this pull request is excellent! I am really looking forward to merging this code and having enhanced QUIC measurement capabilities.

Thank you so much, you did a great job! I'm going to merge as soon as checks are green!

🐳

Issue #1115, opened 2020-12-05
dnscheck: configure using JSON as input


bassosimone:
This is about changing dnscheck to allow forcing http3. While there, it makes sense to reason about way in which we can configure an experiment by providing a JSON document as input rather than just the URL to measure.

This new interface looks a lot like OONI run. What OONI run does for us is that it encodes a JSON with information regarding what to do into the query string. Here we will have code that takes this JSON to do something.

Specifically, we need this kind of flexibility for urlgetter, to make sure we precompute what we want to measure.

In the future it would be nice for this code to be extended to other experiments. It would also be nice for the backend to provide to us information regarding what we need to do. For the time being, let us only implement dnscheck.

bassosimone:
This is ~done! I am going to close this issue with more details before the end of today.

PR #1131, opened 2020-12-10
feat: introduce the experimental run experiment

bassosimone:
This experiment takes in input one or more JSON documents. Each
JSON document describes an experiment that should be run.

The document also allows you to specify all the settings for the
experiment. We currently only support dnscheck.

This is an example of input file:

```JSON
{"name": "dnscheck", "input": "https://dns.google/dns-query", "dns_check":{"http3_enabled": true, "domain": "x.org"}}
{"name": "dnscheck", "input": "dot://dns.google", "dns_check":{"domain": "x.org"}}
```

This runs dnscheck twice. The first time using Google's DoH over
http3, for x.org. The second time using Google's DoT.

Part of https://github.com/ooni/probe-engine/issues/1115.

### DIFF
--- a/experiment/dnscheck/dnscheck.go
+++ b/experiment/dnscheck/dnscheck.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	testName      = "dnscheck"
-	testVersion   = "0.1.0"
+	testVersion   = "0.2.0"
 	defaultDomain = "example.org"
 )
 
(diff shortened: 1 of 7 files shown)


experiment/dnscheck/dnscheck.go
```diff
@@ -22,7 +22,7 @@ import (
 
 const (
 	testName      = "dnscheck"
-	testVersion   = "0.1.0"
+	testVersion   = "0.2.0"
```
bassosimone, line 25: Forgot to bump version in when I last changed it a couple of commits ago.

PR #1138, opened 2020-12-13
build automatically for rpi3 and rpi4

xhdix:
In some cases we need to download the latest file for Raspberry Pi. (If it does not cause side effects.)


bassosimone:
I am going to take a look at this PR in the next Sprint cycle.

bassosimone (commented):
Thank you! I think it's very important to provide binaries for ARM! So, thank you for spearheading this process and prodding us to do that! :-)

I think we should always build for Raspberry Pi when we build for Linux. I've proposed my changes to implement that and pushed them as part of this pull request.

Please, let me know if you like the changes, @xhdix!

bassosimone:
Update at the end of Sprint 30: we're finishing discussing this topic. I'm going to move the issue to the next Sprint. Should we merge earlier than that, we will have some work already done before even starting the Sprint :-).

.github/workflows/miniooni.yml
```diff
@@ -23,6 +23,14 @@ jobs:
...
           name: miniooni-darwin-amd64
           path: ./CLI/darwin/amd64/miniooni
+      - uses: actions/upload-artifact@v1
+        with:
+          name: miniooni-linux-armv7
+          path: ./CLI/linux/armv7/miniooni
+      - uses: actions/upload-artifact@v1
+        with:
+          name: miniooni-linux-arm64
+          path: ./CLI/linux/arm64/miniooni
```
xhdix, line 33: You're welcome! 
Everything is fine, but shouldn't this change be added to the Linux section?
Thanks.

xhdix (commented):
LGTM, only a small change may be needed.

.github/workflows/miniooni.yml
bassosimone: yes, totally! I will indeed move it!

bassosimone:
Too fast

PR #1154, opened 2020-12-18
fix(session): much better code to lookup location

bassosimone:
Reference issue: https://github.com/ooni/probe-engine/issues/1174.


bassosimone:
Before merging, we need to update tests.

Issue #1148, opened 2020-12-16
probeservices: implement checkin API


lorenzoPrimi:
xRef: https://ams-pg-test.ooni.org/apidocs/#/default/post_api_v1_check_in

PR #1158, opened 2020-12-22
probeservices, oonimkall: implement checkin api

lorenzoPrimi:
Fixes https://github.com/ooni/probe-engine/issues/1148


probeservices/checkin.go
```diff
@@ -0,0 +1,24 @@
...
+	V     int               `json:"v"`
+}
+
+// CheckIn pobes ask for tests to be run or otherwise go back to sleep
+// The config argument contains the mandatory settings.
+// Returns the list of tests to run and the URLs, on success, or an explanatory error, in case of failure.
+func (c Client) CheckIn(ctx context.Context, config model.CheckInConfig) (model.CheckInInfo, error) {
+	var response checkInResult
+	err := c.Client.PostJSON(ctx, "/api/v1/check-in", config, &response)
+	if err != nil {
```
bassosimone, line 20:
```suggestion
	if err := c.Client.PostJSON(ctx, "/api/v1/check-in", config, &response); err != nil {
```

model/checkininfo.go
```diff
@@ -0,0 +1,11 @@
+package model
+
+// CheckInInfoWebConnectivity contains the array of URLs returned by the checkin API
+type CheckInInfoWebConnectivity struct {
+	URLs []URLInfo `json:"urls"`
```
bassosimone, line 5: IIRC there is also the reportID we should be using. I would like to add this to the structure even though we are currently not using it?

probeservices/checkin_test.go
```diff
@@ -0,0 +1,66 @@
...
+	client := newclient()
+	client.BaseURL = "https://ams-pg-test.ooni.org"
+	config := model.CheckInConfig{
+		Charging:        true,
+		OnWiFi:          true,
+		Platform:        "string",
+		ProbeASN:        "string",
+		ProbeCC:         "string",
+		RunType:         "string",
+		SoftwareVersion: "string",
```
bassosimone, line 21: Can we use realistic values here?

probeservices/checkin_test.go
```diff
@@ -0,0 +1,66 @@
...
+		WebConnectivity: model.CategoryCodes{
+			CategoryCodes: []string{"NEWS", "CULTR"},
+		},
+	}
+	ctx := context.Background()
+	result, err := client.CheckIn(ctx, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	//TODO what do we want to check?
```
bassosimone, line 31:
```suggestion
	// TODO(lorenzoPrimi): what do we want to check?
```

bassosimone: Apart from this style-oriented comment, I think checking that the category codes are okay is already good. I also think we should be checking whether we have got a report ID.

model/checkininfo.go
```diff
@@ -0,0 +1,11 @@
...
+}
+
+// CheckInInfo contains the return test objects from the checkin API
+type CheckInInfo struct {
+	WebConnectivity CheckInInfoWebConnectivity `json:"web_connectivity"`
```
bassosimone, line 10: My understanding is that this property is optional. If that is the case, then we should use instead:

```suggestion
	WebConnectivity *CheckInInfoWebConnectivity `json:"web_connectivity"`
```

This also means that we need to be careful when manipulating the returned value. This is a sad side effect of languages that allow for `nil` values and we need to live with that.

probeservices/checkin.go
```diff
@@ -0,0 +1,24 @@
...
+
+type checkInResult struct {
+	Tests model.CheckInInfo `json:"tests"`
+	V     int               `json:"v"`
+}
+
+// CheckIn pobes ask for tests to be run or otherwise go back to sleep
+// The config argument contains the mandatory settings.
+// Returns the list of tests to run and the URLs, on success, or an explanatory error, in case of failure.
+func (c Client) CheckIn(ctx context.Context, config model.CheckInConfig) (model.CheckInInfo, error) {
```
bassosimone, line 17:
```suggestion
func (c Client) CheckIn(ctx context.Context, config model.CheckInConfig) (*model.CheckInInfo, error) {
```

I think we want to return either a valid pointer or an error. This is a minor thing, but I actually happen to quite like this either-or semantics where we basically emulate the returning of a sum type.

After this change, tests need to be updated.

model/checkinconfig.go
```diff
@@ -0,0 +1,18 @@
+package model
+
+// CategoryCodes is an array of category codes
+type CategoryCodes struct {
+	CategoryCodes []string `json:"category_codes"` // CategoryCodes is an array of category codes
+}
```
bassosimone, line 6: Maybe I'm wrong, but I thought we already had a structure for doing this? If we do not, then I suggest it may be a good idea to call this structure `CheckInCategoryCodes`.

probeservices/checkin.go
bassosimone: I like this style more, because it's more compact.

probeservices/checkin_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if len(result.WebConnectivity.URLs) < 1 {
```
bassosimone, line 32: This needs to check for `result.WebConnectivity == nil` once we make this field nullable.

probeservices/checkin_test.go
```diff
@@ -0,0 +1,66 @@
...
+			CategoryCodes: []string{"NEWS", "CULTR"},
+		},
+	}
+	ctx := context.Background()
+	result, err := client.CheckIn(ctx, config)
+	if err == nil || !strings.HasSuffix(err.Error(), "invalid control character in URL") {
+		t.Fatal("not the error we expected")
+	}
+	//TODO how to do a proper check here?
+	if len(result.WebConnectivity.URLs) != 0 {
```
bassosimone, line 63: On failure, we want `result` to be `nil`. I think there is one extra case to test where the server returns us a valid but empty WebConnectivity. We can either provoke this case using the real server or using mocking.

bassosimone (commented):
Thank you for working on this! This code is already a very robust base to apply small incremental changes on! I have provided suggestions on how to adapt the API a little bit and make the tests more robust.

probeservices/checkin_test.go
lorenzoPrimi: I don't understand this report_id for what is needed, and we are not gonna use it anytime soon. Do we need the test now?

probeservices/checkin_test.go
lorenzoPrimi: Platform:        "android",
		ProbeASN:        "AS12353",
		ProbeCC:         "PT",
		RunType:         "?",
		SoftwareVersion: "2.7.1",

lorenzoPrimi: I don't know what runtype is. @FedericoCeratto ?

model/checkinconfig.go
lorenzoPrimi: There is no other CategoryCodes struct

probeservices/checkin_test.go
lorenzoPrimi: solved

model/checkinconfig.go
bassosimone: So, I suggest either you put it into `model/categorycodes.go` or prefix `CheckIn` to it?

bassosimone: Now that I take another look at the code, I believe we should just have `model/checkin.go` with everything inside it using the `CheckIn` prefix to make it super easy to understand where the code lives!

model/checkininfo.go
bassosimone: Let us put all in `model/checkin.go` so this can become `CheckInWebConnectivity`!

probeservices/checkin.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/ooni/probe-engine/model"
+)
+
+type checkInResult struct {
+	Tests model.CheckInInfo `json:"tests"`
+	V     int               `json:"v"`
+}
+
+// CheckIn pobes ask for tests to be run or otherwise go back to sleep
```
bassosimone, line 14: I think we need to spell check this sentence!

bassosimone (commented):
Did another pass!

bassosimone:
Remaining TODOs:

- [ ] make sure we generate the desired interfaces for Android/iOS
- [ ] re-read the whole diff
- [ ] make sure we're testing everything we should be testing

bassosimone (approved):
:whale:

Issue #1125, opened 2020-12-08
Routine releases in Sprint 30


bassosimone:
- [x] psiphon: run ./update.bash
- [x] engine: update dependencies
- [x] engine: update internal/httpheader/useragent.go
- [x] engine: update version/version.go
- [x] engine: update resources/assets.go
- [x] engine: update bundled certs (using `go generate ./...`)
- [x] engine: make sure all workflows are green
- [x] engine: tag a new version
- [x] engine: update again version.go to be alpha
- [x] engine: create release at GitHub
- [x] engine: update mobile-staging branch to create oonimkall
- [x] cli: pin to latest engine
- [x] cli: update internal/version/version.go
- [x] cli: tag a new version
- [x] cli: update internal/version/version.go again to be alpha
- [x] android: pin to latest oonimkall
- [x] ios: pin to latest oonimkall
- [x] desktop: pin to latest cli
- [x] engine: create issue for next routine release
- [x] e2etesting: see whether we can remove legacy checks

bassosimone:
We have an issue where we cannot build for Android: https://github.com/ooni/probe-engine/runs/1691128694?check_suite_focus=true.

bassosimone:
done!

PR #1180, opened 2021-01-12
fix(android): pin ndk version and repair build

bassosimone:
While there, ensure we build Android and iOS also in release
branches. So we notice issues _before_ releasing.


bassosimone:
Follow-up of https://github.com/ooni/probe-engine/issues/1125. We're fixing the Android build.

PR #1182, opened 2021-01-13
extended run for urlgetter

kelmenhorst:

PR #1183, opened 2021-01-14
fix: allow mixing different nettests inside run

bassosimone:
Since https://github.com/ooni/probe-engine/pull/1182 we can mix
urlgetter and dnscheck when using run. This is not awesome because
it violates implicit constraints in pipeline, explorer.

Specifically, a single reportID should always contain just a
single kind of nettests.

On top of that, submitting measurements using a reportID containing
the `run` experiment name was also very lame.

With this diff, we are making a decisive step forward by basically
changing the reportID we're using whenever we need it.

To this end, we change the way in which we submit. We're not using
the algorithm used by mobile resubmission.

This feels like part of https://github.com/ooni/probe-engine/issues/921.


internal/mockable/mockable.go
```diff
@@ -33,6 +33,7 @@ type Session struct {
 	MockableResolverIP           string
 	MockableSoftwareName         string
 	MockableSoftwareVersion      string
+	MockableSubmitter            model.Submitter
```
bassosimone, line 36: TODO(@bassosimone): we should either remove this part of the diff or finish upgrading the mocks.

Issue #1120, opened 2020-12-07
session_integration_test.go:309: unexpected KibiBytesSent


bassosimone:
Another CI failure. Occurred to me when running tests locally while working on https://github.com/ooni/probe-engine/issues/1043. Output excerpt:

```
--- FAIL: TestSessionLocationLookup (0.22s)
    session_integration_test.go:309: unexpected KibiBytesSent
```

bassosimone:
Relapse: https://github.com/ooni/probe-engine/runs/1676013667

bassosimone:
Relapse: https://github.com/ooni/probe-engine/pull/1180/checks?check_run_id=1691688358

bassosimone:
Relapse: https://github.com/ooni/probe-engine/pull/1188/checks?check_run_id=1721347958

bassosimone:
This issue is now becoming a high priority because we see it popping up very often.

PR #1188, opened 2021-01-16
chore: bless 0.23.0-alpha.1

bassosimone:


bassosimone:
We have a https://github.com/ooni/probe-engine/issues/1120 failure in the alltests build. Apart from that all green.

PR #1190, opened 2021-01-18
build-cli.sh: support windows/386 and linux/386

bassosimone:
Written while working on https://github.com/ooni/probe/issues/1315


bassosimone:
The build is failing because the test machine disk is full. We'll fix this tomorrow.

Issue #1184, opened 2021-01-14
quic/h3: quality assurance (2/n)


bassosimone:
- [ ] (2021-01-11) take a look a TODOs in the QUIC codebase and add them here
- [ ] DoH should not use h3 unless explicitly asked to (https://github.com/ooni/probe-engine/issues/1113)
- [ ] log transport protocol information in http logging debug-mode-only functionality
- [ ] disable compression by default (so we see all headers)?
- [ ] anything else?

bassosimone:
Moving at beginning of next Sprint.

bassosimone:
Move to next Sprint, since I need more time for https://github.com/ooni/probe/issues/985

PR #1197, opened 2021-01-26
(fix) quic saver: store TLS handshake info

kelmenhorst:
We wrongfully didn't store the TLS information (such as peer certificates, cipher suite) for QUIC. This is fixed by this PR.


bassosimone (approved):
🐳

bassosimone:
Xref: https://github.com/ooni/probe-engine/issues/1184

